PR #3, opened 2024-02-02
[k8s.contrib] and [k8s.contrib.appEnvCluster] Add new packages

holzensp:
Adding the `convert.pkl` script in `k8s.contrib` and a demonstrator of the hierarchy pattern in `k8s.contrib.appEnvCluster`.


stackoverflow (approved):
LGTM

PR #13, opened 2024-02-08
Fix typo

berkus:
Also make docs match docs of allOf.


bioball:
Thanks! You'll also need to bump the package version by one patch version.

bioball:
Rolled into https://github.com/apple/pkl-pantry/pull/34

PR #16, opened 2024-02-18
Fix duplicated class names

valerio-iachini:
This PR fixes a bug that leads to the generation of duplicated codes when multiple JSON schema definitions share the same property name. As a result, the code generation process produces multiple classes whit the same name.

### DIFF
--- a/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
+++ b/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
@@ -222,13 +222,14 @@ function isClassLike(schema: JsonSchema.Schema): Boolean =
 /// Try to use the parent property's name as part of the class name in case of conflict.
 /// If already at the root, add a number at the end.
 local function determineTypeName(path: List<String>, candidateName: String, existingTypeNames: Set<Type>, index: Int): Type =
-  if (existingTypeNames.contains(utils.pascalCase(candidateName)))
-    if (path.isEmpty)
-      determineTypeName(path, candidateName + index.toString(), existingTypeNames, index + 1)
+  let (candidateType = new Type { name = utils.pascalCase(candidateName); moduleName = module.moduleName })
+    if (existingTypeNames.contains(candidateType))
+      if (path.isEmpty)
+        determineTypeName(path, candidateName + index.toString(), existingTypeNames, index + 1)
+      else
+        determineTypeName(path.dropLast(1), utils.pascalCase(path.last.capitalize()) + utils.pascalCase(candidateName), existingTypeNames, index)
     else
-      determineTypeName(path.dropLast(1), utils.pascalCase(path.last.capitalize()) + utils.pascalCase(candidateName), existingTypeNames, index)
-  else
-    new { name = utils.pascalCase(candidateName); moduleName = module.moduleName }
+      candidateType 
 
 /// The schemas that should be rendered as classes.
 ///
@@ -240,7 +241,7 @@ local classSchemas: Type.TypeNames =
       .fold(Map(), (accumulator: Type.TypeNames, pair) ->
         let (path = pair.first)
         let (schema = pair.second)
-        let (typeName = determineTypeName(path, path.lastOrNull?.capitalize() ?? "Item", accumulator.values.toSet(), 0))
+        let (typeName = determineTypeName(path.dropLast(1), path.lastOrNull?.capitalize() ?? "Item", accumulator.values.toSet(), 0))
           accumulator.put(schema, typeName)
       )
 
--- a/packages/org.json_schema.contrib/PklProject
+++ b/packages/org.json_schema.contrib/PklProject
@@ -27,5 +27,5 @@ dependencies {
 }
 
 package {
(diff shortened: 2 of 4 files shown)


packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
```diff
@@ -222,13 +222,14 @@ function isClassLike(schema: JsonSchema.Schema): Boolean =
 /// Try to use the parent property's name as part of the class name in case of conflict.
 /// If already at the root, add a number at the end.
 local function determineTypeName(path: List<String>, candidateName: String, existingTypeNames: Set<Type>, index: Int): Type =
-  if (existingTypeNames.contains(utils.pascalCase(candidateName)))
-    if (path.isEmpty)
-      determineTypeName(path, candidateName + index.toString(), existingTypeNames, index + 1)
+  let (candidateType = new Type { name = utils.pascalCase(candidateName); moduleName = module.moduleName })
+    if (existingTypeNames.contains(candidateType))
```
holzensp, line 226: This can't be right, can it?
```suggestion
    if (existingTypeNames.contains(candidateType.name))
```

Could you add a test for the case that you're fixing?

holzensp (changes requested):
Thanks for your contribution. I _believe_ I spotted a bug. If I'm right, could you make sure that bug is covered in the tests?

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
valerio-iachini: Yes, I'll add tests. `existingTypeNames` contains a list of `Type`, so if we call `contains` with only the name, it always returns `false`. The `determineTypeName` function returns a `Type` that is added to the accumulator, which is passed again to the function. The name `typeNames` is also used because there is an alias [here](https://github.com/apple/pkl-pantry/blob/4f8f939c35ff1fdc04b0897db4b53c05ff234186/packages/org.json_schema.contrib/internal/Type.pkl#L25) that states that a map of JSON schema and `Type` is called `TypesName`.

valerio-iachini: The code that I fixed with this PR incorrectly called the `contains` method with just the name, causing it to consistently return `false`. This resulted in the generation of classes with the same name when encountering properties of type object with identical names in the schema.

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
```diff
@@ -240,7 +241,7 @@ local classSchemas: Type.TypeNames =
       .fold(Map(), (accumulator: Type.TypeNames, pair) ->
         let (path = pair.first)
         let (schema = pair.second)
-        let (typeName = determineTypeName(path, path.lastOrNull?.capitalize() ?? "Item", accumulator.values.toSet(), 0))
+        let (typeName = determineTypeName(path.dropLast(1), path.lastOrNull?.capitalize() ?? "Item", accumulator.values.toSet(), 0))
```
tbinna, line 244: Could you explain why this change is necessary?

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
tbinna: The implementation here may be a more straightforward solution: https://github.com/apple/pkl-pantry/pull/21/files

valerio-iachini: Yeah, I see what you mean, but it's not exactly the same. Is it acceptable to have a type with the same name in a different module? Also, every time the function is called, it remaps all the types to their corresponding names. Additionally, this PR addresses another issue: the generated classes have the property name repeated twice.

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
valerio-iachini: Otherwise, the generated classes end up with the property name repeated twice. You can check out the code snippet [here](https://github.com/apple/pkl-pantry/blob/be26d4171c97b5dca365a90a6397ff2739c5933c/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl#L230C42-L230C52), where the first argument is appended with the current candidate name.

tbinna: Ah yes, I can see this in my output now too. Thanks for clarifying

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
holzensp: You are right; I didn't follow through on the actual definition. In this case, please rename `s/existingTypeNames/exstingTypes/`, because it's pretty misleading currently. I don't want you to be blocked on that, though, because it's PR-scope-creep.

holzensp (approved):
Happy to take this as is (when CI is green); happy for you to roll in the rename.

tbinna:
Fixes #20

Issue #20, opened 2024-02-20
[JSON Schema Contrib] Schema generator produces duplicated class types


tbinna:
I attempted to generate a pkl schema from a JSON schema as [explained here](https://github.com/apple/pkl/issues/92#issuecomment-1936593897). Strangely, the output contained duplicated class definitions, which caused the pkl eval process to fail with

```shell
–– Pkl Error ––
Duplicate definition of member `Target`.

544 | class Target {
      ^^^^^^^^^^^^
```

## Reproduction
Generate a pkl schema from the JSON schema:

```shell
 pkl eval package://pkg.pkl-lang.org/pkl-pantry/org.json_schema.contrib@1.0.0#/generate.pkl -m . -p source="https://raw.githubusercontent.com/toolsplus/pkl-forge/main/out/manifest-schema.json"
```

then run

```shell
pkl eval ManifestSchema.pkl
```

I believe the issue is L225 in the snippet below. `existingTypeNames` has type `Set<Type>`, and the contains function compares the set entries to a string, which always returns false.

https://github.com/apple/pkl-pantry/blob/4f8f939c35ff1fdc04b0897db4b53c05ff234186/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl#L224-L245

## Solution

Update L225 as follows:

```pkl
if (existingTypeNames.map((type) -> type.name).contains(utils.pascalCase(candidateName)))
```

I tested this successfully with the reproduction case.

jstrachan:
can this be closed now?

holzensp:
Interesting; PR was marked by OP. Closing this, assuming that's correct.

PR #21, opened 2024-02-20
[org.json_schema.contrib] Fix duplicated class types

tbinna:
Fix comparison between `existingTypeNames` set and pascal case candidate name string

Resolves: #20


tbinna:
Duplicates #16

PR #28, opened 2024-03-04
Fix package versions

bioball:
Looks like we forgot to bump versions for packages k8s.contrib.crd and pkl.experimental.syntax


bioball:
These changes were folded into #29. Closing this one.

PR #33, opened 2024-03-05
fix(org.json_schema): add Mapping to JsonSchemaValue

### LINKED ISSUE #32: [K8s Contrib CRD] CRD generation failed when a field uses `{}` as default value.
CRD generation will fail when a field uses `{}` as default value.

```
–– Pkl Error ––
Expected value of type `Boolean|Number|Null|String|Dynamic|Typed|Listing`, but got type `Mapping`.
Value: new Mapping {}

39 | typealias JsonSchemaValue = Boolean|Number|Null|String|Dynamic|Typed|Listing
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at org.json_schema.JsonSchema#default (projectpackage://pkg.pkl-lang.org/pkl-pantry/org.json_schema@1.0.0#/JsonSchema.pkl)

101 | default = raw.getOrNull("default")
                ^^^^^^^^^^^^^^^^^^^^^^^^
at org.json_schema.Parser#toJsonSchema.default (projectpackage://pkg.pkl-lang.org/pkl-pantry/org.json_schema@1.0.0#/Parser.pkl)

......
```

A minimal CRD to reproduce the problem:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: example
spec:
  scope: Namespaced
  group: example
  names:
    kind: Example
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          properties:
            some_map:
              additionalProperties:
                type: string
              default: {}
              type: object
          type: object
```

inoc603:
Closes #32

### DIFF
--- a/packages/k8s.contrib.crd/PklProject
+++ b/packages/k8s.contrib.crd/PklProject
@@ -32,5 +32,5 @@ dependencies {
 }
 
 package {
-  version = "1.0.1"
+  version = "1.0.2"
 }
--- a/packages/k8s.contrib.crd/PklProject.deps.json
+++ b/packages/k8s.contrib.crd/PklProject.deps.json
@@ -29,12 +29,12 @@
     },
     "package://pkg.pkl-lang.org/pkl-pantry/org.json_schema.contrib@1": {
       "type": "local",
-      "uri": "projectpackage://pkg.pkl-lang.org/pkl-pantry/org.json_schema.contrib@1.0.2",
+      "uri": "projectpackage://pkg.pkl-lang.org/pkl-pantry/org.json_schema.contrib@1.0.3",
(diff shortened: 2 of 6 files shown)


inoc603:
I tested manually with the example in #32 , but I didn't find any documentation on how to write tests for these packages, and I haven't add any test yet. Although I can kind of guess how to do it based on existing ones, I would hope there's a guide for writing tests for pkl.

bioball:
Thanks!

I am going to roll this change into #34 so that it can be put into the same release.

PR #37, opened 2024-03-06
[com.circleci.v2] Improve Config module

bioball:
* Add tests and examples
* Add OrbStep method for easy interaction with step definitions coming from an orb
* Export class `AbstracStep` so it can be subclassed to define typed orb definitions


stackoverflow (approved):
LGTM

PR #43, opened 2024-03-24
Add env render module

jtagrgh:
Hi,

This PR addresses https://github.com/apple/pkl/issues/273. 

I assumed the .env format defined by https://github.com/motdotla/dotenv.

Since .env can only represent single line value assignments (e.g., NAME="foo") the primary challenge was deciding how to simplify pkl structures. Here, I decided to minimize parsing for the .env consumer.

So, I chose the following conversions:
- Hierarchical structures (Mappings) are flattened, using keys as the components.
- Array structures (Listings) are flattened, using indexes as the components.

So,
```
// .pkl file
name="joe"
server {
  host = "localhost"
  port = 8080
}
```
becomes
```
// .env file
NAME="joe"
SERVER_HOST="localhost"
SERVER_PORT="8080"
```

And
```
// .pkl file
ports {
  8080
  8081
  8082
}
```
becomes
```
// .env file
PORTS_0="8080"
PORTS_1="8081"
PORTS_2="8082"
```

Alternatively, array structures could be inlined, so that the above example would instead become
```
PORTS="[8080, 8081, 8082]"
```
letting flattening be the responsibility of converters.

Thanks for reading!


bioball:
Hey, thanks for the contribution!

This seems a little too opinionated for something like pkl-pantry--the handling of nested values seems a little too tied to the dotenv library, and doesn't seem generally usable as a format. For that reason, I don't think pkl-pantry is the best fit for this.

However, you don't need to submit this package here in order for it to be shareable. Feel free to publish it yourself as a GitHub release. When you do that, you can import the package via `package://pkg.pkl-lang.org/github.com/<org>/<repo>/<release>@<version>`.

For an example of another package that's published independently, take a look at @jamesward's https://github.com/jamesward/pklgha

PR #46, opened 2024-04-02
[pkl.experimental.net] Disallow unicode digits in IPs, ports

lilyball:
This PR introduces a test suite for IP address strings, complete with failing tests for IP address strings containing unicode digits, and then fixes the failing tests by changing the regexes to stop allowing unicode digits.

As part of this work I also converted the IPv6 regex into expanded form with whitespaces and comments in order to make it more readable so I could verify the test suite covered everything. The new regex is identical to the old one so no behavior changed there. That commit could be removed without changing the rest of this PR but it seems nice to have.

### DIFF
--- a/packages/pkl.experimental.net/PklProject
+++ b/packages/pkl.experimental.net/PklProject
@@ -17,5 +17,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.1.0"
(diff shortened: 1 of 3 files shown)


HT154 (approved):
Awesome!

bioball (approved):
LGTM, thank you!

PR #49, opened 2024-04-04
[pkl.experimental.uri] Zero-pad percent encoded values

rnapier:


holzensp (changes requested):
Sorry, trigger-happy; you need to bump the version [here](https://github.com/apple/pkl-pantry/blob/ce6b87c0568636290bb86b624bcd137d367de511/packages/pkl.experimental.uri/PklProject#L19), or the package can't publish.

bioball:
Huh. I was trying to fix this PR, but that ended up auto-closing it with a force-push. That was weird; but re-submitting the same fix in https://github.com/apple/pkl-pantry/pull/54

PR #51, opened 2024-04-11
Add error output about which dependent tasks need version bumps

bioball:
This adds an error message to the `createPackages` task, to print names of dependent packages that also need their package version bumped.

Sample:

```
> Package `package://pkg.pkl-lang.org/pkl-pantry/pkl.experimental.uri@1.0.0` was already published with different contents.

  Computed checksum: e115ec6942c5169817953bca73c2349cee5e11981aa4db68b9f1e7aed616394c
  Published checksum: 12a42da6a2933a802cc79cea7f5541513b5106070ca5f1236009ebefeb3d81b3

  This likely means that you need to bump the packages version.
> The following packages depend on this package, and also need to have their version updated:
  * k8s.contrib.crd
  * org.json_schema.contrib
  * org.json_schema
```

Also: make Gradle track inputs for project resolution. It means that `resolveProjects` is considered up to date if no `PklProject`s have changed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,10 +1,16 @@
+import groovy.json.JsonSlurper
+import org.jetbrains.kotlin.utils.sure
 import org.pkl.core.Version
+import org.pkl.gradle.task.ProjectPackageTask
 import java.io.OutputStream
 import java.net.URI
 import java.nio.file.Files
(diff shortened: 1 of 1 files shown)


holzensp (approved):
Nice

PR #52, opened 2024-04-12
[org.openapis.v3.contrib] Omit `required` when no fields are actually required.

mrstegeman:
Generating an empty `required` list is a spec violation.


bioball (approved):
LGTM

PR #53, opened 2024-04-12
[pkl.lua] New package

lilyball:
`pkl.lua` is a package with one module, `pkl.lua.lua`, which provides a `ValueRenderer` subclass that serializes to Lua.


lilyball:
Is there any way to test pkldoc generation before publishing? I can't seem to get the command-line version working locally.

lilyball:
Thanks that worked great! Though I also had to bump the pkl version in `gradle/libs.versions.toml` to get some documentation render fixes.

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+    value.replaceAllMapped(Regex(#"[\p{Cntrl}\\\#(delim)&&[^ ]]"#), (match) ->
+      if (match.value == "\u{7}") "\\a"
+      else if (match.value == "\u{8}") "\\b"
+      else if (match.value == "\u{12}") "\\f"
+      else if (match.value == "\n") if (multiline) "\\\n" else "\\n"
+      else if (match.value == "\r") "\\r"
+      else if (match.value == "\t") "\\t"
+      else if (match.value == "\u{11}") "\\v"
+      else if (match.value == delim) "\\\(delim)"
+      else "\\u{\(match.value.codepoints.first.toRadixString(16))}")
```
bioball, line 232: This would be a runtime error, so I guess this isn't being tested right now.

```suggestion
      else "\\u{\(match.value.codePoints.first.toRadixString(16))}")
```

Also, `codePoints.first` is a little suspicious. Do you need to worry about surrogate pairs here?

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+    let (multiline = map.length >= multilineThreshold && level != null && value is Mapping)
+    let (level_ = if (multiline) level + 1 else level)
+    (if (multiline) "{\n\(indent.repeat(level_))" else "{ ")
+    + new Listing {
+      for (k,v in map) {
+        let (path = path?.add(Key(k)))
+        "\(renderKey(k, "")) = \(render(convert(v, path), path, level_));"
+      }
+    }.join(if (multiline) "\n\(indent.repeat(level_))" else " ")
+    + if (multiline) "\n\(indent.repeat(level))}" else " }"
```
bioball, line 251:
```suggestion
  local function renderMap(value: Mapping|Map, path: List<PathEntry>?, level: UInt?): String =
    if (value.isEmpty) "{}"
    else
      let (map = value.toMap())
      let (multiline = map.length >= multilineThreshold && level != null && value is Mapping)
      let (level_ = if (multiline) level!! + 1 else level)
        (if (multiline) "{\n\(indent.repeat(level_!!))" else "{ ")
        + new Listing {
          for (k,v in map) {
            let (path = path?.add(Key(k)))
            "\(renderKey(k, "")) = \(render(convert(v, path), path, level_));"
          }
        }.join(if (multiline) "\n\(indent.repeat(level_!!))" else " ")
        + if (multiline) "\n\(indent.repeat(level!!))}" else " }"
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+  local function renderList(value: Listing|List, path: List<PathEntry>?, level: UInt?): String =
+    if (value.isEmpty) "{}" else
+    let (path = path?.add(splatKey))
+    let (multiline = value.length >= multilineThreshold && level != null && !(value is List))
+    let (level_ = if (multiline) level + 1 else level)
+    (if (multiline) "{\n\(indent.repeat(level_))" else "{ ")
+    + value.toList()
+        .map((it) -> render(convert(it, path), path, level_))
+        .join(if (multiline) ",\n\(indent.repeat(level_))" else ", ")
+    + if (multiline) "\n\(indent.repeat(level))}" else " }"
```
bioball, line 262:
```suggestion
  local function renderList(value: Listing|List, path: List<PathEntry>?, level: UInt?): String =
    if (value.isEmpty) "{}" else
      let (path = path?.add(splatKey))
      let (multiline = value.length >= multilineThreshold && level != null && !(value is List))
      let (level_ = if (multiline) level!! + 1 else level)
        (if (multiline) "{\n\(indent.repeat(level_!!))" else "{ ")
        + value.toList()
            .map((it) -> render(convert(it, path), path, level_))
            .join(if (multiline) ",\n\(indent.repeat(level_!!))" else ", ")
        + if (multiline) "\n\(indent.repeat(level!!))}" else " }"
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+      for (i,elt in list) {
+        when (i < list.lastIndex) {
+          "\(render(convert(elt, listPath), listPath, level_)),"
+        } else {
+          render(convert(list.last, listPath), listPath, level_)
+        }
+      }
+    }.join(if (multiline) "\n\(indent.repeat(level_))" else " ")
+    + if (multiline) "\n\(indent.repeat(level))}" else " }"
+}
```
bioball, line 292:
```suggestion
  local function renderDynamic(value: Dynamic|Typed, path: List<PathEntry>?, level: UInt?): String =
    let (list = if (value is Dynamic) value.toList() else List())
    let (map = value.toMap())
    // note: Map.keys.toList() is O(1), other ways of converting are O(n) (as of Pkl 0.25.3)
    let (entries = map.keys.toList().mapNonNull((k) ->
      let (path = path?.add(Prop(k)))
      let (v = convert(map[k], path))
        if (omitNullProperties && v == null && k is String) null
        else Pair(k, Pair(path, v))
    ))
      if (entries.isEmpty && list.isEmpty) "{}"
      else
        let (multiline = entries.length + list.length >= multilineThreshold && level != null)
        let (level_ = if (multiline) level!! + 1 else level)
        let (listPath = if (list.isEmpty) null else path?.add(splatKey))
          (if (multiline) "{\n\(indent.repeat(level_!!))" else "{ ")
          + new Listing {
            for (kpv in entries) { // kpv = Pair(key, Pair(path, value))
              "\(renderKey(kpv.key, "")) = \(render(kpv.value.value, kpv.value.key, level_));"
            }
            for (i,elt in list) {
              when (i < list.lastIndex) {
                "\(render(convert(elt, listPath), listPath, level_)),"
              } else {
                render(convert(list.last, listPath), listPath, level_)
              }
            }
          }.join(if (multiline) "\n\(indent.repeat(level_!!))" else " ")
          + if (multiline) "\n\(indent.repeat(level!!))}" else " }"
}
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+    converterMap.any((k,_) -> k is Class && reflect.Class(k).isSubclassOf(typedClass))
+
+  local function convert(value: Any, path: List<PathEntry>?): Any =
+    if (path == null) value else
+    let (path = path.reverse())
+    let (f = pathConverters.findOrNull((p) -> comparePaths(path, p.key))?.value)
+    let (klass = value.getClass())
+    let (f = f ?? converterMap.getOrNull(klass))
+    let (f = f ?? if (hasTypedConverters && value is Typed) findTypedConverter(reflect.Class(klass).superclass) else null) // find superclass converters
+    if (f != null) f.apply(value) else value
```
bioball, line 169:
```suggestion
  local function convert(value: Any, path: List<PathEntry>?): Any =
    if (path == null) value
    else
      let (path = path.reverse())
      let (f = pathConverters.findOrNull((p) -> comparePaths(path, p.key))?.value)
      let (klass = value.getClass())
      let (f = f ?? converterMap.getOrNull(klass))
      let (f = f ?? if (hasTypedConverters && value is Typed) findTypedConverter(reflect.Class(klass).superclass) else null) // find superclass converters
        if (f != null) f.apply(value) else value
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+    else throw("Converter path `\(key)` has invalid syntax."))
+  .entries
+
+  // [true] if the converters define any class converters.
+  // For the time being this is limited to any subclasses of [Typed], as this matches current
+  // [PcfRenderer] behavior. I understand this is a Pkl bug and e.g. [Number] should match [Int],
+  // but we'll match the bug for now for performance reasons.
+  local hasTypedConverters: Boolean =
+    let (typedClass = reflect.Class(Typed))
+    converterMap.any((k,_) -> k is Class && reflect.Class(k).isSubclassOf(typedClass))
```
bioball, line 160:
```suggestion
    let (typedClass = reflect.Class(Typed))
      converterMap.any((k,_) -> k is Class && reflect.Class(k).isSubclassOf(typedClass))
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+    if (key.matches(pathSpecRegex))
+      if (key == "") List("^") else // PcfRenderer treats empty path as "^"
+      key.split(pathSpecSplitRegex)
+      .map((it) ->
+        if (it == "^") it
+        else if (it.startsWith("[")) new Key { key = it.drop(1).dropLast(1) }
+        else new Prop { name = it })
+      .reverse()
+    else throw("Converter path `\(key)` has invalid syntax."))
+  .entries
```
bioball, line 152:
```suggestion
  // noinspection TypeMismatch
  local pathConverters: List<Pair<List<PathEntry>,(unknown) -> Any>> =
    converterMap
      .filter((key, _) -> key is String)
      .mapKeys((key, _) ->
        if (key.matches(pathSpecRegex))
          if (key == "") List("^")
          else // PcfRenderer treats empty path as "^"
            key
              .split(pathSpecSplitRegex)
              .map((it) ->
                if (it == "^") it
                else if (it.startsWith("[")) new Key { key = it.drop(1).dropLast(1) }
                else new Prop { name = it })
              .reverse()
        else throw("Converter path `\(key)` has invalid syntax."))
      .entries
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+
+  // ## Misc helpers ##
+
+  // Returns a [Prop] for string keys, otherwise [splatKey].
+  local function Prop(k: Any): Prop|Key =
+    if (k is String) new Prop { name = k } else splatKey
+
+  // Returns a [Key] for string keys, otherwise [splatKey]
+  local function Key(k: Any): Key =
+    if (k is String) new Key { key = k } else splatKey
```
bioball, line 133: [nit] place these functions above their class definitions.

E.g.

```
// Returns a [Prop] for string keys, otherwise [splatKey].
local const function Prop(k: Any): Prop|Key =
  if (k is String) new Prop { name = k } else splatKey

local class Prop {
  name: String
}
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+
+  // ## Converters ##
```
bioball, line 135: [nit] use `region...endregion` comments.

Add this here
```suggestion
  // region Converters
```

Then add `// endregion` to mark the end of the region.

See `JsonSchema.pkl` for an example

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+  ///
+  /// For more information see [ValueRenderer.converters]. Note that due to language limitations,
+  /// any entries in a [Dynamic] will have converters applied as though the entry was a property.
+  converters: Mapping<(Class|String), (unknown) -> Any>
+
+  extension = "lua"
+
+  function renderValue(value: Any): String =
+    let (path = List("^"))
+    render(convert(value, path), path, 0)
```
bioball, line 100:
```suggestion
      render(convert(value, path), path, 0)
```

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,292 @@
...
+        let (path = path.add(splatKey))
+        let (v = convert(v, path))
+        // remember, Lua indexes are 1-based
+        acc + "_ENV[\(idx+1)] = \(render(v, path, 0))\n")
+    else if (value is Mapping|Map)
+      value.fold("", (acc, k, v) ->
+        let (path = path.add(Key(k)))
+        let (v = convert(v, path))
+        acc + "\(renderKey(k, "_ENV")) = \(render(v, path, 0))\n")
+    else throw("The top-level value of a Lua document must have type `Typed`, `Dynamic`, `Mapping`, or `Map`, but got type `\(value.getClass())`")
```
bioball, line 123:
```suggestion
  function renderDocument(value: Any): String =
    let (path = List("^"))
    let (value = convert(value, path))
      if (value is RenderDirective) "\(value.text)\n"
      else if (value is LuaRenderDirective) "\(value.before ?? "")\(render(value.value, path, 0))\(value.after ?? "")\n"
      else if (value is Dynamic|Typed)
        value.toMap().fold("", (acc, k, v) ->
          let (path = path.add(Prop(k)))
          let (v = convert(v, path))
            if (omitNullProperties && v == null && k is String) acc
            else acc + "\(renderKey(k, "_ENV")) = \(render(v, path, 0))\n")
        + (if (value is Dynamic) value.toList() else List()).foldIndexed("", (idx, acc, v) ->
          let (path = path.add(splatKey))
          let (v = convert(v, path))
          // remember, Lua indexes are 1-based
            acc + "_ENV[\(idx+1)] = \(render(v, path, 0))\n")
      else if (value is Mapping|Map)
        value.fold("", (acc, k, v) ->
          let (path = path.add(Key(k)))
          let (v = convert(v, path))
            acc + "\(renderKey(k, "_ENV")) = \(render(v, path, 0))\n")
      else throw("The top-level value of a Lua document must have type `Typed`, `Dynamic`, `Mapping`, or `Map`, but got type `\(value.getClass())`")
```

packages/pkl.lua/tests/lua.pkl
```diff
@@ -0,0 +1,367 @@
...
+      }
+    }.renderDocument(source) == new lua.Renderer {}.renderDocument(expected)
+    // check for string vs int key on Dynamic too; note that Dynamic treats all string keys as properties, so the converter shouldn't run
+    let (source = new Dynamic { [1] = 1; ["1"] = 1; nest { [1] = 1; ["1"] = 1 } })
+    let (expected = new Dynamic { [1] = 1; ["1"] = 1; nest { [1] = 1; ["1"] = 1 } })
+    new lua.Renderer {
+      converters {
+        ["[1]"] = (it) -> it + 1
+      }
+    }.renderDocument(source) == new lua.Renderer {}.renderDocument(expected)
```
bioball, line 211: This case isn't working right now:

```groovy
import "@lua/lua"

foo {
  ["bar"] = 5
}

output {
  renderer = new lua.Renderer {
    converters {
      ["foo[bar]"] = (it) -> it + 1
    }
  }
}
```

I would expect this to produce:

```
foo = { bar = 6; }
```

bioball: Also, you might want to consider writing these as examples, rather than facts, because they help you diff output when things fail.

bioball (changes requested):
This is looking pretty good! Have mostly style comments, but some other things that need looking at.

packages/pkl.lua/lua.pkl
```diff
@@ -0,0 +1,283 @@
...
+/// A string that is a Lua reserved keyword.
+///
+/// These strings are not allowed to be used as identifiers.
+typealias LuaKeyword = "and"|"break"|"do"|"else"|"elseif"|"end"|"false"|"for"|"function"|"goto"|"if"|"in"|"local"|"nil"|"not"|"or"|"repeat"|"return"|"then"|"true"|"until"|"while"
+
+/// A string that is a valid Lua identifier.
+typealias LuaIdentifier = String(matches(Regex("[a-zA-Z_][a-zA-Z0-9_]*")) && !(this is LuaKeyword))
+
+/// Directs [Renderer] to output additional text [before] and/or [after] rendering a [value].
+typealias LuaRenderDirective = PcfRenderDirective
```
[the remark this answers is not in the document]
lilyball: I ended up replacing it with its own class already. And the intended use-case here is to be able to emit code that operates on values instead of just emitting values. For example, you could have a converter like `[MyClass] = (it) -> new LuaRenderDirective { before = "setmetatable(", value = it, after = ", classMetatable)" }`. Such a converter would cause `MyClass` to be emitted as a table with a particular metatable attached.

packages/pkl.lua/lua.pkl
lilyball: Is this documented anywhere? The VSCode extension doesn't seem to treat this as folding markers. I'm happy to do this though I'd like to understand its purpose.

packages/pkl.lua/tests/lua.pkl
lilyball: This is a known failure case, it's due to a language limitation where Pkl code cannot distinguish between properties and entries when iterating over a `Dynamic`. This limitation is mentioned in the doc comment on the `converters` property (and is the reason why that property is redeclared).

packages/pkl.lua/lua.pkl
lilyball: Good catch! I'll make sure there's a test that exercises this code path.

packages/pkl.lua/lua.pkl
lilyball: What does `// noinspection TypeMismatch` do?

lilyball:
I have applied all suggestions, with the sole exception that where you recommended doing
```pkl
if (condition) value
else
  …
```

I left it as
```pkl
if (condition) value else
  …
```

One of your suggestions actually used the latter style, and there are a few other instances of this style in the repository, and the style guide does not cover this case, so I figured it was fine.

I also fixed the indentation in a couple other spots that were not mentioned.

Curiously, moving `Prop()` and `Key()` out of the class definition required me to add the `const` modifier and I don't know why that would be necessary.

packages/pkl.lua/lua.pkl
bioball: Yup, they're folding markers. Only the IJ plugin understands these right now, but we should add support for this in our LSP so this behavior is consistent across editors.

It's not documented anywhere at the moment.

packages/pkl.lua/tests/lua.pkl
bioball: Gotcha, sounds good!

bioball:
> One of your suggestions actually used the latter style, and there are a few other instances of this style in the repository, and the style guide does not cover this case, so I figured it was fine.

I guess you're right that we don't cover this. But, I find this much easier to read, because it's clearer that it belongs to the `else` block.

```
if (condition) value
else
  …
```

> Curiously, moving Prop() and Key() out of the class definition required me to add the const modifier and I don't know why that would be necessary.

This is because class bodies can only reference outside `const` members. Otherwise, an amended module can possibly change the class definition.

Consider:

```groovy
// base.pkl
bar = "bar"

function foo() = bar

class MyClass {
  myProp = foo()
}
```

Then:

```groovy
amends "base.pkl"

bar = "barbar"
```

lilyball:
Ahh right, I was thinking here that the function definition itself couldn't be replaced, but I didn't think about the function being able to reference other non-`const` values from the module.

packages/pkl.lua/lua.pkl
lilyball: Nevermind, I just checked out the IntelliJ plugin and now I understand.

bioball (approved):
LGTM. Good work!

PR #55, opened 2024-04-24
[pkl.experimental.net] IPv6 fixes including IPv4-suffixes, scope ids, address formatting

lilyball:
This includes a lot of changes to the IPv6 handling of this package. Highlights include:

- The alternative IPv4-suffixed format is now supported for all addresses, not just IPv4-mapped addresses.
- Scope IDs are now supported for all non-global addresses, not just link-local unicast.
- `expandIPv6AddressString()` now understands the alternative IPv4-suffixed format and scope IDs.
- `compressIPv6AddressString()` now understands the alternative IPv4-suffixed format and scope IDs, and its output conforms to [RFC5952 §4.2.3](https://datatracker.ietf.org/doc/html/rfc5952#section-4.2.3).
- `IPv6Address()` now understands the alternative IPv4-suffixed format and scope IDs.
- `IPv6Address` now has a `scopeId` field, `toString()` now produces fully canonicalized addresses, and a separate method `toExpandedString()` has been added for use-cases that require the fully expanded string.
- `UInt128.words` is no longer `hidden`, allowing `==` to work correctly on it.
- `IPv4Address.repr` is no longer `hidden`, allowing `==` to work correctly on it.
- `IPv6Address.repr` is no longer `hidden`, allowing `==` to work correctly on it.
- `IPNetwork()`/`IPv4Network()`/`IPv6Network()` now use correct typing on their parameter.
- `IPv6Network` understands `scopeId`.
- `IPv6Range()` preserves `scopeId` when both addresses have the same one.


bioball (approved):
Great work!

PR #58, opened 2024-05-14
[com.circleci] Add missing resource classes, fix job constraint

bioball:


stackoverflow (approved):

packages/com.circleci.v2/PklProject
```diff
@@ -17,5 +17,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.1.1"
+  version = "1.1.2"
```
holzensp, line 20: nit; in the strictest sense, isn't this `1.2.0`?

holzensp (approved):

packages/com.circleci.v2/PklProject
bioball: It's a bugfix though?

PR #60, opened 2024-05-28
[pkl.lua] Add `lua.Parser`

lilyball:
This updates the `pkl.lua` package to have a `Parser` class that can parse a static subset of Lua. It's targeting the same subset of Lua that `Renderer` outputs (namely, a document consisting of key/value pairs consisting entirely of literals, with no expressions), though it recognizes all Lua syntax, including handling all forms of comments and literals that `Renderer` does not emit (such as hexadecimal floating point literals).

The parser's interface is modeled after `pkl.json#Parser`, though it uses the property `useDynamic` instead of `useMapping` as Lua syntax does not make a distinction between objects and arrays. Like `pkl.json#Parser`, path converters are run as though all string keys are properties, even when configured to return a `Mapping`.

In addition to adding the parser, this PR also updates `Renderer` to start applying converters to mapping entry keys, to match `PcfRenderer`'s behavior. Any mapping entry key[^1] has class converters applied to it, and if the key is itself an object, nested values will be converted the same way `PcfRenderer` does (which is to say, any nested values have a path that starts at the object and omits the `"^"` marker, so e.g. a key `[new Dynamic { a = 1 }]` has a single nested value whose full path is `"a"`).

This also renames `LuaIdentifier`, `LuaKeyword`, and `LuaRenderDirective` to drop the leading "Lua", though it leaves deprecated typealiases behind for compatibility.

This PR also bumps the project version to 1.1.0.

[^1]: Reminder that due to a language limitation, when rendering a `Dynamic`, any mapping entries with a `String` key are indistinguishable from properties.


packages/pkl.lua/lua.pkl
```diff
@@ -294,3 +332,560 @@ class Renderer extends ValueRenderer {
...
+
+local const byteLut = new Mapping {
+  for (k,v in nybbleLut) {
+    for (k2,v2 in nybbleLut) {
+      ["\(k)\(k2)"] = v.shl(4) + v2
+    }
+  }
+}
+
+// endregion
```
lilyball, line 889: These functions were based on some functions found in [pkl.experimental.net/net.pkl](https://github.com/apple/pkl-pantry/blob/437359005c0c4d2afdac2e2fb191740d3dc0a073/packages/pkl.experimental.net/net.pkl#L498-L515). The changes made here compared to the originals were done for performance reasons, as benchmarking showed that a lookup table is significantly faster than pkl.experimental.net's approach for `parseHex()`, and expanding the table to include uppercase letters is faster than lowercasing the input.

I am tempted to propose that we make a new package `pkl.hex` that contains these functions along with any other useful hex manipulation utilities, although it would probably be better just to update the Pkl stdlib to support parsing with radix natively.

packages/pkl.lua/lua.pkl
```diff
@@ -294,3 +332,560 @@ class Renderer extends ValueRenderer {
...
+    --(?:               # comment
+      \[(?<eq1>=*)\[    # long comment (equals are GROUP 1)
+      (?>(?>
+        [^]]
+      | ](?!\k<eq1>])
+      )*)
+      (]\k<eq1>]|\z)    # match end or EOF (GROUP 2)
+    | .*                # short comment
+    )
+  | [\w&&\D]\w*         # identifier
```
bioball, line 755: `\D` is any character that is not a digit. This is probably not what you meant?  This matches `*foo` for example.

packages/pkl.lua/lua.pkl
```diff
@@ -294,3 +332,560 @@ class Renderer extends ValueRenderer {
...
+    | ](?!\k<eq2>])
+    )*)
+    (]\k<eq2>]|\z)      # match end or EOF (GROUP 6)
+    # for the numeric literals, they consume extra periods and hex digits and throw a parse error
+    # this regex aims to match everything Lua 5.3 itself will tokenize as numeric
+  | \.?0[xX](?:(?:[pP][-+]?)?[.\p{XDigit}])*   # hex numeric literal
+  | \.?\d(?:(?:[eE][-+]?)?[.\p{XDigit}])*   # dec numeric literal
+  | [-+*%^\#&|(){}\[\];,]   # single-char operators
+  | <[<=]? | >[>=]? | //? | ~=? | ==? | ::? | \.{1,3}   # multi-char operators
+  | ([^\ \f\n\t\x0b])   # invalid token (last group)
```
bioball, line 776: [nit] unnecessary escapes here (found this using IJ)

```suggestion
    | \\(?>z[ \f\n\t\x0b]*|x\p{XDigit}{2}|\d{1,3}|u\{\p{XDigit}+}|(?s:.)|\z)
    )*)("|(?s:.)|\z)    # match string truncated by newline or EOF (GROUP 3)
  | '(?>(?>             # short literal string (')
      [^\\\n']
    | \\(?>z[ \f\n\t\x0b]*|x\p{XDigit}{2}|\d{1,3}|u\{\p{XDigit}+}|(?s:.)|\z)
    )*)('|(?s:.)|\z)    # match string truncated by newline or EOF (GROUP 4)
  | \[(?<eq2>=*)\[      # long literal string (equals are GROUP 5)
    (?>(?>
      [^]]
    | ](?!\k<eq2>])
    )*)
    (]\k<eq2>]|\z)      # match end or EOF (GROUP 6)
    # for the numeric literals, they consume extra periods and hex digits and throw a parse error
    # this regex aims to match everything Lua 5.3 itself will tokenize as numeric
  | \.?0[xX](?:(?:[pP][-+]?)?[.\p{XDigit}])*   # hex numeric literal
  | \.?\d(?:(?:[eE][-+]?)?[.\p{XDigit}])*   # dec numeric literal
  | [-+*%^\#&|(){}\[\];,]   # single-char operators
  | <[<=]? | >[>=]? | //? | ~=? | ==? | ::? | \.{1,3}   # multi-char operators
  | ([^ \f\n\t\x0b])   # invalid token (last group)
```

packages/pkl.lua/lua.pkl
bioball: Yeah, we already have `Int#toRadixString`. I think it makes sense to have `String#toRadixInt()` as well, which also complements `String#toInt()`.

packages/pkl.lua/lua.pkl
```diff
@@ -294,3 +332,560 @@ class Renderer extends ValueRenderer {
...
+
+// checks the error states documented on tokenRegex
+// returns the same token, or throws an error
+local const function validateToken(token: RegexMatch, source: String, uri: Uri?): RegexMatch =
+  if (token.groups[2]?.value == "") throwError("expected ]\(token.groups[1]!!.value)], found EOF", source, uri, token.groups[2]!!)
+  else let (g = token.groups[3]) if (g != null && g.value != "\"") throwExpected("\"", source, uri, g)
+  else let (g = token.groups[4]) if (g != null && g.value != "'") throwExpected("'", source, uri, g)
+  else if (token.groups[6]?.value == "") throwError("expected ]\(token.groups[5]!!.value)], found EOF", source, uri, token.groups[6]!!)
+  else let (g = token.groups.last) if (g != null) throwError("illegal token \(g.value)", source, uri, g)
+  else token
```
bioball, line 787: For consistency sake: error messages should start with a capitalized word.

```suggestion
  if (token.groups[2]?.value == "") throwError("Expected ]\(token.groups[1]!!.value)], found EOF", source, uri, token.groups[2]!!)
  else let (g = token.groups[3]) if (g != null && g.value != "\"") throwExpected("\"", source, uri, g)
  else let (g = token.groups[4]) if (g != null && g.value != "'") throwExpected("'", source, uri, g)
  else if (token.groups[6]?.value == "") throwError("Expected ]\(token.groups[5]!!.value)], found EOF", source, uri, token.groups[6]!!)
  else let (g = token.groups.last) if (g != null) throwError("Illegal token \(g.value)", source, uri, g)
  else token
```

packages/pkl.lua/lua.pkl
```diff
@@ -294,3 +332,560 @@ class Renderer extends ValueRenderer {
...
+  op: ("["|"key"|"]"|"="|"value")? // op is "["|"value"? iff key is null
+  mapStart: RegexMatch? // non-null if !map.isEmpty
+  listStart: RegexMatch? // non-null if !list.isEmpty
+
+  function toValue(useDynamic: Boolean, source: String, uri: Uri?): Dynamic|Mapping|Listing =
+    if (useDynamic) (map.toDynamic()) {
+      ...list
+    } else if (map.isEmpty) list.toListing()
+    else if (list.isEmpty) map.toMapping()
+    else throwError2("table has both list elements and map entries", source, uri, mapStart!!, "first map entry", listStart!!, "first list entry")
```
bioball, line 701:
```suggestion
    else throwError2("Table has both list elements and map entries", source, uri, mapStart!!, "first map entry", listStart!!, "first list entry")
```

packages/pkl.lua/tests/parser.pkl-expected.pcf
```diff
@@ -0,0 +1,232 @@
...
+        if (k is Dynamic) new PcfRenderDirective { before = "Dynamic "; value = k }
+        else if (k is Mapping) new PcfRenderDirective { before = "Mapping "; value = k }
+        else if (k is Listing) new PcfRenderDirective { before = "Listing "; value = k }
+        else k)
+        if (v is Dynamic|Mapping|Listing) Pair(key, fixupTableKeys(v))
+        else Pair(key,v)
+    ))
+      if (value is Dynamic) (map.toDynamic()) {
+        ...value.toList()
+      } else map.toMapping()
```
bioball, line 232: While I see the use-case, this isn't valid PCF. This happens to work right now, but it will break when we start using an actual PCF parser for running tests.

I think the best way to write these tests, for now, is to use the PCF renderer for generating examples.

If/when we have an in-language Pkl renderer, that will make these tests even better because you will get output like `foo = new Dynamic {}`.

packages/pkl.lua/tests/parser.pkl-expected.pcf
```diff
@@ -0,0 +1,232 @@
...
+      | ^
+    at <input>:1:1
+    """.replaceAll("\n", " ")
+    """
+    _ENV cannot be assigned to directly
+
+    1 | _ENV=1
+      |     ^
+    at <input>:1:5
+    """.replaceAll("\n", " ")
```
bioball, line 127: Very cool! Great job with the error reporting here; this looks great.

Per my other comment, we can't have method calls here; this is not valid PCF.

I think let's make these facts, and do a string equality in your tests.

For example:

```groovy
facts {
  ["errors"] {
    module.catch(() -> parser.parse("_ENV=1")) == """
      _ENV cannot be assigned to directly

      1 | _ENV=1
        |     ^
      at <input>:1:5
      """.replaceAll("\n", " ")
  }
}
```

bioball (changes requested):
Well done! This is looking really good.

I have some nits, and some blocking comments about the tests.

packages/pkl.lua/lua.pkl
lilyball: `[\w&&\D]` means the intersection of `\w` and `\D`, so the pattern `[\w&&\D]\w*` means "a word character that's not a digit, followed by zero or more word characters". Since I've disabled unicode character classes, this is equivalent to writing `[a-zA-Z_][a-zA-Z0-9_]*`.

I originally had this as the definition for `identifierRegex` too, but I changed it there to get rid of a false-positive warning claiming that `[\w&&\D]` can be simplified to `\w&&\D`. Perhaps I should have changed it here too for consistency.

packages/pkl.lua/tests/parser.pkl-expected.pcf
lilyball: This code here is necessary precisely because the Pcf renderer produces output that cannot be parsed. As long as the tests all pass it's fine, but as soon as a test fails then Pkl needs to be able to parse the `parser.pkl-actual.pcf` file. This means I need to muck with the output in the test file, which causes it to render something that is parseable, but then it fails the equality comparison (since it's now comparing a value that includes `PcfRenderDirective` with something that doesn't). So I need the same code in the pcf file to modify the expected value to have the same `PcfRenderDirective` values.

I suppose I could do a different hack that replaces the Dynamic used as a key with e.g. a string containing the rendered Pcf. Or I could just write the `PcfRenderDirective` values inline in the output, but then that raises the possibility of `fixupTableKeys()` producing ordering differences in the output.

Is this a blocker?

packages/pkl.lua/lua.pkl
lilyball: Done

packages/pkl.lua/tests/parser.pkl-expected.pcf
lilyball: Done. The string approach worked fine.

packages/pkl.lua/tests/parser.pkl-expected.pcf
lilyball: Done

lilyball:
Force-push was to fix an indentation issue in the new commits.

lilyball:
@bioball I've addressed your feedback, though I left the `[\w&&\D]\w*` intact in the source as expanding that to `[a-zA-Z_][a-zA-Z0-9_]*` negatively affected formatting of the regex comments.

packages/pkl.lua/lua.pkl
bioball: Ah, I see; read this as a character class at first. My bad!

bioball (approved):
Amazing stuff, great work!

PR #68, opened 2024-07-01
[pkl.experimental.syntax] Do not generate spaces at EOL for long typealiases

HT154:
There's some extraneous diff in `packages/pkl.experimental.syntax/tests/ExpressionNode.pkl-expected.pcf` because I ran `pkl test --overwrite` and it subbed in raw strings where there were plain strings with escapes previously. I figure this is probably desirable overall, but I'd be happy to revert it based on feedback.


holzensp (approved):
LGTM

PR #69, opened 2024-07-04
[org.json_schema.contrib] Support `allOf` schemas

HT154:
Lots of limitations here still, module doc comment for generate.pkl has been updated accordingly.

This PR applies IntelliJ formatting and `pkl test --overwrite`, so I recommend hiding whitespace diffs during review.

### DIFF
--- a/packages/org.json_schema.contrib/internal/TypesGenerator.pkl
+++ b/packages/org.json_schema.contrib/internal/TypesGenerator.pkl
@@ -454,7 +454,8 @@ local function generateNumberType(schema: JsonSchema): Pair<JsonSchema, TypeNode
     if (isInt)
       let (minimum = if (schema.exclusiveMinimum != null) schema.exclusiveMinimum!! + 1 else schema.minimum)
       let (maximum = if (schema.exclusiveMaximum != null) schema.exclusiveMaximum!! - 1 else schema.maximum)
-      let (refinedInt = if (minimum == 0 && maximum == math.maxUInt32)
+      let (refinedInt =
+        if (minimum == 0 && maximum == math.maxUInt32)
           utils.declaredType("UInt32")
         else if (minimum == 0 && maximum == math.maxUInt16)
           utils.declaredType("UInt16")
@@ -685,25 +686,33 @@ function collateAllOf(schema: JsonSchema(allOf != null)): JsonSchema =
           // metadata
           title = collateInformation(elem.title, res.title)
           description = collateInformation(elem.description, res.description)
-          default = if (elem.default != null && res.default != null) let (_ = trace("Unable to combine allOf elements into one schema: dropping conflicting default")) res.default
+          default =
+            if (elem.default != null && res.default != null) let (_ = trace("Unable to combine allOf elements into one schema: dropping conflicting default")) res.default
             else elem.default ?? res.default
-          examples = if (elem.examples != null && res.examples != null) new {
-            when (elem.examples is Listing) { ...elem.examples as Listing } else { elem.examples }
-            when (res.examples is Listing) { ...res.examples as Listing } else { res.examples }
-          } else elem.examples ?? res.examples
-          deprecated = if (elem.deprecated == true && res.deprecated == true) true
+          examples =
+            if (elem.examples != null && res.examples != null)
+              new {
+                when (elem.examples is Listing) { ...elem.examples as Listing } else { elem.examples }
+                when (res.examples is Listing) { ...res.examples as Listing } else { res.examples }
+              }
+            else elem.examples ?? res.examples
+          deprecated =
+            if (elem.deprecated == true && res.deprecated == true) true
             else null // if we have a mix of null/false/true, this is undeterminable
           readOnly = collateMetadataBoolean(elem.readOnly, res.readOnly, "readOnly")
           writeOnly = collateMetadataBoolean(elem.writeOnly, res.writeOnly, "writeOnly")
 
           // core
-          type = if (res.type != null && elem.type != null && res.type != elem.type) allOfErr("conflicting type")
-          else elem.type ?? res.type
-          `const` = if (res.`const` != null && elem.`const` != null && res.`const` != elem.`const`) allOfErr("conflicting const")
-          else elem.`const` ?? res.`const`
+          type =
+            if (res.type != null && elem.type != null && res.type != elem.type) allOfErr("conflicting type")
+            else elem.type ?? res.type
+          `const` =
+            if (res.`const` != null && elem.`const` != null && res.`const` != elem.`const`) allOfErr("conflicting const")
+            else elem.`const` ?? res.`const`
 
           // number
-          multipleOf = if (elem.multipleOf != null && res.multipleOf != null)
+          multipleOf =
+            if (elem.multipleOf != null && res.multipleOf != null)
             if (elem.multipleOf is Int && res.multipleOf is Int) math.lcm(elem.multipleOf!! as Int, res.multipleOf!! as Int)
             else allOfErr("multiple non-integer multipleOf")
           else elem.multipleOf ?? res.multipleOf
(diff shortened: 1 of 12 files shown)


packages/org.json_schema.contrib/internal/TypesGenerator.pkl
```diff
@@ -63,6 +63,9 @@ function generateTypeNode(
   else if (typeNames.containsKey(schema))
     let (type = typeNames[schema as JsonSchema])
       utils.declaredType1(type, type.moduleName != enclosingModuleName)
+  else if (typeNames.containsKey((schema) { type = null }))
+    let (type = typeNames[(schema) { type = null } as JsonSchema])
+      utils.declaredType1(type, type.moduleName != enclosingModuleName)
```
HT154, line 68: Ran into a fun case with input that looks like this:
```json
{
  "type": "object",
  "$ref": "#/$defs/users_groups.user"
}
```

Where the `schema` in the lookup `typeNames.containsKey(schema)` has `type = "object"` but the actual dereferenced schema has `type = null`. This causes a fall through to `Any` and pretty unhelpful code generation!

This particular fix is a workaround for what seems to be a common use case for `$ref`, but this doesn't handle the general case of arbitrary schema fields being set on the referrer side. I'd be very happy to take suggestions on how to handle this generally.

packages/org.json_schema.contrib/internal/TypesGenerator.pkl
```diff
@@ -60,8 +60,8 @@ function generateTypeNode(
     else
       new TypeNode.BuiltInTypeNode { type = "nothing" }
   // If we have generated a class or typealias definition already, simply use it.
-  else if (typeNames.containsKey(schema))
-    let (type = typeNames[schema as JsonSchema])
+  else if (typeNames.containsKey(schema._inline_?.getOrNull("__ref_orig__") ?? schema))
+    let (type = typeNames[(schema._inline_?.getOrNull("__ref_orig__") ?? schema) as JsonSchema])
```
HT154, line 64: Ran into a fun case with input that looks like this:
```json
{
  "type": "object",
  "$ref": "#/$defs/users_groups.user"
}
```

Where the `schema` in the lookup `typeNames.containsKey(schema)` has `type = "object"` but the actual dereferenced schema has `type = null`. This causes a fall through to `Any` and pretty unhelpful code generation!

`ref.resolveRef` now sets `_inline_["__ref_orig__"] in its output to facilitate this lookup.

packages/org.json_schema.contrib/ref.pkl
```diff
@@ -27,8 +27,8 @@ local function parseJsonPointer(ref: String): List<String|Int> =
   ref
     .split("/")
     .map((str) -> str
-      .replaceAll("~0", "~")
-      .replaceAll("~1", "/"))
+    .replaceAll("~0", "~")
+    .replaceAll("~1", "/"))
```
bioball, line 31: Revert; this method chain is within the `map()` so the indent is intentional

packages/org.json_schema.contrib/internal/TypesGenerator.pkl
```diff
@@ -669,5 +669,100 @@ local function generateBaseType(schema: JsonSchema, typeNames: Type.TypeNames):
...
+            if (elem.multipleOf is Int && res.multipleOf is Int) math.lcm(elem.multipleOf!! as Int, res.multipleOf!! as Int)
+            else allOfErr("multiple non-integer multipleOf")
+          else elem.multipleOf ?? res.multipleOf
+          minimum = collateMin(elem.minimum, res.minimum)
+          exclusiveMinimum = collateMin(elem.exclusiveMinimum, res.exclusiveMinimum)
+          maximum = collateMax(elem.maximum, res.maximum)
+          exclusiveMaximum = collateMax(elem.exclusiveMaximum, res.exclusiveMaximum)
+
+          // string
+          pattern = if (elem.pattern != null && res.pattern != null) "\(elem.pattern)|\(res.pattern)"
```
bioball, line 702: Probably best to wrap each alternative in a non-capture group so each sub-pattern is treated as its own alternative.

```suggestion
          pattern = if (elem.pattern != null && res.pattern != null) "(?:\(elem.pattern))|(?:\(res.pattern))"
```

packages/org.json_schema.contrib/internal/TypesGenerator.pkl
```diff
@@ -669,5 +669,100 @@ local function generateBaseType(schema: JsonSchema, typeNames: Type.TypeNames):
...
+          additionalItems = collateSchema(elem.additionalItems, res.additionalItems, "additionalItems")
+          minItems = collateMin(elem.minItems, res.minItems) as UInt?
+          maxItems = collateMax(elem.maxItems, res.maxItems) as UInt?
+          uniqueItems = if (elem.uniqueItems == true || res.uniqueItems == true) true
+            else elem.uniqueItems ?? res.uniqueItems
+
+          // composition
+          oneOf = collateSet(elem.oneOf, res.oneOf)
+          anyOf = collateSet(elem.anyOf, res.anyOf)
+          not = collateSchema(elem.not, res.not, "not")
```
bioball, line 733: There's some things that are missing that would be good to collate:

* title
* description
* default
* deprecated
* examples
* readOnly
* writeOnly

bioball (changes requested):
Overall, great work!!

Left some minor comments.

Also, one issue I'm noticing here is that it can't successfully generate Pkl for schemas whose root definition only contains `allOf`, and no `properties`.

This patch fixes that for me:

```patch
diff --git a/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl b/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
index 84ef091..6c0a6eb 100644
--- a/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
+++ b/packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
@@ -37,6 +37,8 @@ local jsonRenderer = new JsonRenderer {}
 /// The root schema, used to resolve `$ref` values.
 rootSchema: JsonSchema
 
+local collatedRootSchema = TypesGenerator.collateAllOf(rootSchema)
+
 /// The URI representing the root schema, used to resolve `$ref` values.
 baseUri: URI
 
@@ -74,7 +76,7 @@ moduleNode: ModuleNode =
         }
       }
       declaration {
-        docComment = getDocComment(rootSchema, "module")
+        docComment = getDocComment(collatedRootSchema, "module")
         when (moduleName != null) {
           moduleHeader {
             name {
@@ -88,8 +90,8 @@ moduleNode: ModuleNode =
         }
       }
       properties =
-        if (isClassLike(rootSchema))
-          generateClassBody(rootSchema, allTypeNames)
+        if (isClassLike(collatedRootSchema))
+          generateClassBody(collatedRootSchema, allTypeNames)
         else null
     }
 
@@ -220,7 +222,7 @@ local function determineTypeName(path: List<String>, candidateName: String, exis
 ///
 /// Classes get rendered for any subschema that has [JsonSchema.properties] defined, and does not have [JsonSchema.`$ref`] defined.
 local classSchemas: Type.TypeNames =
-  let (schemas = utils._findMatchingSubSchemas(rootSchema, List(), (elem) -> elem != rootSchema && isClassLike(elem)))
+  let (schemas = utils._findMatchingSubSchemas(collatedRootSchema, List(), (elem) -> elem != collatedRootSchema && isClassLike(elem)))
     schemas
       .entries
       .fold(Map(), (accumulator: Type.TypeNames, pair) ->
@@ -238,7 +240,7 @@ local classNames: Set<Type> = classSchemas.values.toSet()
 /// Typealiases get rendered for any [JsonSchema.definitions] or [JsonSchema.`$defs`] that should not be rendered as a class.
 local typeAliasSchemas: Type.TypeNames =
   // Grab all schemas that have `definitions` or `$defs`
-  let (schemasWithDefinitions = utils._findMatchingSubSchemas(rootSchema, List(), (elem) -> (elem.definitions ?? elem.$defs) != null))
+  let (schemasWithDefinitions = utils._findMatchingSubSchemas(collatedRootSchema, List(), (elem) -> (elem.definitions ?? elem.$defs) != null))
     schemasWithDefinitions
       .entries
     // For each schema, return the child json schema properties that are not class-like
```

For local testing, try out the tsconfig schema:

```
pkl eval generate.pkl -p source=https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/tsconfig.json -m .out
```

packages/org.json_schema.contrib/ref.pkl
bioball: This comment generally applies everywhere else, too. Do you mind undoing the formatting changes? I'm guessing some of these are committed by our auto-formatter, but IMO it makes it a little harder to read (we'll address all of this when we have a formatting tool that ships with Pkl)

HT154: The keyboard shortcut is second nature, sorry 😅 I'll revert the lines I'm not otherwise touching.

packages/org.json_schema.contrib/internal/TypesGenerator.pkl
HT154: How would each of these be collated? I didn't address these because I wasn't sure there was a right way. Here's how I might do these:
* title - join by newlines? commas?
* description - join by newlines
* default - error if more than one value set
* deprecated - I don't think there's a valid way to do this
* examples - if all elements with non-null examples are listings, append them, otherwise error
* readOnly - all sub-schemas must set the same readOnly value? all sub-schemas with non-null readOnly must set the same value?
* writeOnly - same as readOnly

bioball: My suggestions (I'm making this up too, so feel free to push back)

* title, description: maybe join by `\n---\n`? That would create a line break in the markdown doc comment.
* default: makes sense
* examples: concat into a listing
* readOnly, writeOnly: your suggestion makes sense
* deprecated: same treatment as readOnly, writeOnly

HT154: Ended up having to take a slightly different tack for a couple fields:
* default - `trace` a warning when there's a conflict otherwise fall back to the last set subschema's default
* deprecated - only mark the result deprecated if all subschemas are marked deprecated, otherwise null

bioball (approved):
LGTM, great work here!

PR #70, opened 2024-07-06
Bugfix: support empty Mapping in TOML output

freynder:
Extended the test case with an empty `Mapping` value, which did not produce a table heading in the previous version although it should.

pkl should eval an empty `Map` to an output that contains the TOML table header without further values.

The previous version explicitly skipped rendering the header assuming there were deeper tables that would follow, however that is not the case with an empty `Map`


bioball (approved):
Good fix, thank you!

PR #72, opened 2024-07-10
[io.prometheus] Implement many missing properties

HT154:


packages/io.prometheus/Configuration.pkl
```diff
@@ -252,10 +234,14 @@ class KubernetesSdConfig {
...
   selectors: Listing<KubernetesSdConfigSelector>?
+
+  /// Optional metadata to attach to discovered targets.
+  /// If omitted, no additional metadata is attached.
+  attach_metadata: KubernetesSdConfigAttachMetadata(node?.implies(role == "pod" || role == "endpoints" || role == "endpointslice") || true)?
 }
 
 class KubernetesSdConfigSelector {
-  role: *"pod"|"service"|"node"|"endpoints"|"ingress"
+  role: * "pod"|"service"|"node"|"endpoints"|"ingress"
```
bioball, line 244:
```suggestion
  role: *"pod"|"service"|"node"|"endpoints"|"ingress"
```

packages/io.prometheus/Configuration.pkl
```diff
@@ -384,7 +397,7 @@ class RelabelConfig {
 /// * `"labeldrop"`: Match `regex` against all label names.
 ///   Any label that matches will be removed from the set of labels.
 /// * `"labelkeep"`: Match `regex` against all label names.
-typealias RelabelAction = *"replace"|"keep"|"drop"|"hashmod"|"labelmap"|"labeldrop"|"labelkeep"
+typealias RelabelAction = * "replace"|"keep"|"drop"|"hashmod"|"labelmap"|"labeldrop"|"labelkeep"
```
bioball, line 400:
```suggestion
typealias RelabelAction = *"replace"|"keep"|"drop"|"hashmod"|"labelmap"|"labeldrop"|"labelkeep"
```

packages/io.prometheus/Configuration.pkl
```diff
@@ -623,32 +619,70 @@ class RemoteReadConfig {
...
+  authorization: Authorization((basic_auth ?? oauth2) == null)?
 
-  /// Sets the `Authorization` header on every remote read request
-  /// with the bearer token read from the configured file.
-  ///
-  /// Mutually exclusive with [bearer_token].
-  bearer_token_file: String?
+  /// Optional OAuth 2.0 configuration.
+  /// Cannot be used at the same time as basic_auth or authorization.
+  oauth2: OAuth2((basic_auth ?? authorization) == null)?
```
bioball, line 666: This will become a stack overflow during typecheck if multiple of these properties are not null, because the constraints refer to each other.

To avoid this, move the constraint to just one of the properties. You can borrow the `onlyOneSet` method from the CircleCI package:

https://github.com/apple/pkl-pantry/blob/7b1670188f8fabdc253985566b8fa5d77aed09a4/packages/com.circleci.v2/Config.pkl#L657-L676

packages/io.prometheus/Configuration.pkl
```diff
@@ -623,32 +619,70 @@ class RemoteReadConfig {
...
+  /// Use proxy URL indicated by environment variables
+  /// (HTTP_PROXY, https_proxy, HTTPs_PROXY, https_proxy, and no_proxy)
+  proxy_from_environment: Boolean?
 
-  /// Sets the `Authorization` header on every remote read request
+  /// Specifies headers to send to proxies during CONNECT requests.
+  proxy_connect_haeder: Mapping<String, Listing<String>>?
+}
+
+open class HttpClientConfig extends ProxyConfig {
```
bioball, line 656:
```suggestion
abstract class HttpClientConfig extends ProxyConfig {
```

packages/io.prometheus/PklProject
```diff
@@ -18,5 +18,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.1.3"
+  version = "2.0.0"
```
bioball, line 21: Is this a breaking change?

bioball (changes requested):
Looks good! I have some minor comments.

packages/io.prometheus/PklProject
HT154: Yup:
* Removed `bearer_token` and `bearer_token_file` - I could not find reference to these anywhere in the prom docs and this usage is covered by `Authorization`
* `RemoteWriteConfig.headers` was previously of type `Labels?`, I've updated it to `Mapping<LabelNameFormat, String>` as [the corresponding struct field is `map[string]string`](https://github.com/prometheus/prometheus/blob/58963f7b2e0b0ec345e3672d2e048627a8961fb3/config/config.go#L1106)

packages/io.prometheus/Configuration.pkl
HT154: Should these have default cases set at all? The only usage is nullable so it's never being defaulted to `"replace"` meaningfully.

packages/io.prometheus/PklProject
bioball: Ideally, we try to avoid bumping major versions if it can be avoided. Introducing a new major version means that prometheus 1.x can't be reconciled with a 2.x version within a project, which can cause headaches WRT the diamond dependency problem.

For the two addressed breaking changes:

* `bearer_token` and `bearer_token_file` exist in [prometheus 1.8](https://prometheus.io/docs/prometheus/1.8/configuration/configuration/). Looks like they got removed somewhere? But their docs don't mention this anywhere as a breaking change.
* It's actually safe to keep `headers: Labels?`. Prometheus will accept ints/booleans as strings; the YAML `someValue: 5` gets parsed into into `map[string]string{"someValue": "5"}` thanks to how gopkg.in/yaml.v2 works

Can you undo these, and make this a minor version bump instead?

bioball: Looking at this more, I think `bearer_token`, `bearer_token_file` [still exist](https://github.com/prometheus/common/blob/8742f090d9fd2cea7e53a4fd58ae5fb6c6634336/config/http_config.go#L306-L311)

HT154: Huh, not sure how I messed that up. I thought I cmd-F'd that file for `bearer_token` and everything 😖

I've put those properties back, marked deprecated to match the source struct.

bioball (approved):
LGTM!

packages/io.prometheus/Configuration.pkl
```diff
@@ -657,17 +711,85 @@ class BasicAuth {
...
+  /// If unset, Prometheus will use Go default maximum version, which is TLS 1.3.
+  /// See MaxVersion in <https://pkg.go.dev/crypto/tls#Config>.
+  max_version: TLSVersion?
 }
+
+typealias TLSVersion = "TLS10"|"TLS11"|"TLS12"|"TLS13"
+
+/// Tells if only one of the values is not null.
+local const function onlyOneSet(values: List<Any>): Boolean =
+  values.filter((it) -> it != null).length == 1
```
bioball, line 795: Re: test failure: this should probably be `atMostOneSet`

PR #81, opened 2024-10-31
Fix typo in CircleCI's `setup_remote_docker` parameter; s/docker_layer_cacheing/docker_layer_caching/

yyuu:
According to CircleCI documentation, `docker_layer_caching` must be correct spelling of the parameter.

> docker_layer_caching | N | boolean | Set this to true to enable Docker Layer Caching in the Remote Docker Environment (default: false)
> Set this to true to enable [Docker Layer Caching](https://circleci.com/docs/docker-layer-caching/) in the Remote Docker Environment (default: false)

https://circleci.com/docs/configuration-reference/#setupremotedocker


bioball (approved):
Ah, good catch, thanks!

PR #82, opened 2024-11-01
[com.circleci.v2] Make `setup_remote_docker` parameters as optional

yyuu:
According to CircleCI documentation on [setup_remote_docker](https://circleci.com/docs/configuration-reference/#setupremotedocker), both `version` and `docker_layer_caching` are NOT required one.

Key | Required | Type | Description
-- | -- | -- | --
docker_layer_caching | N | boolean | Set this to true to enable Docker Layer Caching in the Remote Docker Environment (default: false)
version | N | String | Version string of Docker you would like to use (default: 24.0.9). View the list of supported Docker versions here.

cf. https://circleci.com/docs/configuration-reference/#setupremotedocker


bioball (approved):
LGTM

PR #83, opened 2024-11-01
[com.circleci.v2] Support job-level parameters

yyuu:
Just as similar as pipeline parameters (toplevel), CircleCI also supports `parameters` at job-level (cf. https://circleci.com/docs/configuration-reference/#parameters-job). The parameter syntax is [identical to the pipeline parameters](https://circleci.com/docs/reusing-config/#parameter-syntax).

Job-level parameters are useful especially when I'm configuring [matrix builds](https://circleci.com/docs/configuration-reference/#matrix). I'd like this to be supported :pray:

### DIFF
--- a/packages/com.circleci.v2/PklProject
+++ b/packages/com.circleci.v2/PklProject
@@ -17,5 +17,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.1.6"
+  version = "1.2.0"
 }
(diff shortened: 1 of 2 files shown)


packages/com.circleci.v2/PklProject
```diff
@@ -17,5 +17,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.1.5"
+  version = "1.1.6"
```
bioball, line 20:
```suggestion
  version = "1.2.0"
```

bioball (approved):

PR #85, opened 2024-12-08
Parameter passing from workflow job

lopezm94:
Allow passing parameters from a workflow job to a job definition as can be seen in circleci documentation [here](https://circleci.com/docs/pipeline-variables/#element-parameter-scope). I'm attempting to use it for generating pipelines to different terraform workspaces dynamically as seen in the text example included.

I wasn't sure what was the more idiomatic way to make these parameters be embedded at the same level as the rest of the params in `WorkflowJob`, so I just added a case on the `YamlRenderer.converters` field. 

In my own code I have a validation to guarantee the fields inside are not repeated with the following code:
```pkl
  parameters: Mapping<String, String|Boolean|Number>?

  // Use reflection to get property names
  hidden reservedFields: Set<String> =
    reflect.Class(CustomWorkflowJob).properties
      .filter((name, _) -> name != "parameters" && name != "reservedFields" && name != "validateExtraFields")
      .mapValues((_, prop) -> prop.name)
      .values
      .toSet()

  hidden validateParameters = !parameters.keys.any((key) -> reservedFields.contains(key)) || throw("Extra fields cannot use reserved names: \(reservedFields)")
```

I didn't add this validation to avoid importing the `reflect` package.

### DIFF
--- a/packages/com.circleci.v2/Config.pkl
+++ b/packages/com.circleci.v2/Config.pkl
@@ -366,6 +366,11 @@ class WorkflowJob {
 
   /// Job Filters can have the key branches or tags
   filters: JobFilters?
+
+  /// A job may have input parameters in their defintion.
+  ///
+  /// A mapping of the parameter name and value must be passed when instatiation the job.
+  parameters: Mapping<String, String|Boolean|Number>?
 }
 
 class JobFilters {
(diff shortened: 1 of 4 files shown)


bioball:
Ah, we got two PRs for the same thing! This is closed by #86

PR #86, opened 2024-12-09
[com.circleci.v2] Support specifying custom job-level parameters while calling a job in a workflow

yyuu:
By default CircleCI workflow jobs can take parameters like `name`, `context`, `type`, `filters`, etc. Along with these default parameters, it is not very explicitly described on [CircleCI's configuration reference](https://circleci.com/docs/configuration-reference/#job-name-in-workflow), however, it also can take custom job-level parameters defined by users.

So far there was no way to specify these custom parameters while calling a workflow job. This allows user to specify custom `parameters` while invoking a WorkflowJob.

---
Here's some references of usage of job-level custom parameters. JFYI

https://github.com/CircleCI-Public/aws-ecr-orb/blob/v9.3.7/src/examples/build_and_push_image_w_registry_login.yml#L21-L36
https://github.com/CircleCI-Public/aws-ecr-orb/blob/v9.3.7/src/jobs/build_and_push_image.yml#L8-L214

### DIFF
--- a/packages/com.circleci.v2/Config.pkl
+++ b/packages/com.circleci.v2/Config.pkl
@@ -367,12 +367,16 @@ class WorkflowJob {
   /// Job Filters can have the key branches or tags
   filters: JobFilters?
 
-  /// Supplemental job-level parameters to be specified while calling a job in workflow.
+  /// Supplemental job-level parameters to be specified when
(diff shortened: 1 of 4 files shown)


bioball (approved):
Thanks!

Polished this up a bit with a test, and also added `Listing<Step>` as a possible type.

PR #87, opened 2024-12-09
[com.circleci.v2] Support CircleCI workflow job matrix

yyuu:
Adds support for CircleCI's matrix jobs.

The configuration schema was simple enough to implement it in Pkl as it is.

Key | Required | Type | Description
-- | -- | -- | --
parameters | Y | Map | A map of parameter names to every value the job should be called with
exclude | N | List | A list of argument maps that should be excluded from the matrix
alias | N | String | An alias for the matrix, usable from another job’s requires stanza. Defaults to the name of the job being executed

https://circleci.com/docs/configuration-reference/#matrix


bioball (approved):
LGTM, thanks!

PR #88, opened 2024-12-12
[pkl.experimental.deepToTyped] Add support for key transformations during apply

HT154:
Sometimes a complex, nested Pkl class/module must be "unmarshalled" from a serialized format like JSON or YAML. `pkl.experimental.deepToTyped` is the low-boilerplate tool of choice for such transformations. Occasionally, the names of properties in the input data do not match the property names of Pkl types, for example Pkl typically uses `camelCase` while JSON can use `snake_case` or `kebab-case`. `deepToTyped` cannot currently handle this task.

This PR introduces a new customization point in the `deepToTyped` module: the `keyTransform` property can be set to a function mapping a `reflect.Property` to the `String` name that should be used to find the property's value in the input instead of its name. If this function returns `null`, the property's original name is used.

### DIFF
--- a/packages/k8s.contrib.crd/PklProject
+++ b/packages/k8s.contrib.crd/PklProject
@@ -29,5 +29,5 @@ dependencies {
 }
 
 package {
-  version = "1.0.13"
+  version = "1.0.14"
 }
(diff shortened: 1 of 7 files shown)


bioball (approved):
Looks good!

Issue #90, opened 2024-12-28
[org.openapis.v3] Apple PKL OpenAPI Security Scheme for HTTP/Bearer Auth cannot be configured correctly


nikeokoronkwo:
This issue addresses the issue of being able to make a correctly configured security scheme for HTTP security schemes like Bearer Auth Tokens.

According to the [OpenAPI Specification](https://spec.openapis.org/oas/v3.0.3.html) (as demonstrated by [Swagger](https://swagger.io/docs/specification/v3_0/authentication/bearer-authentication/)), when specifying a security scheme with `type` as `"http"`, the properties `name` and `in` are not allowed. However, the properties are still required even when the type is set to `"http"` (the properties required are `scheme`, with `bearerFormat` being optional).

https://github.com/apple/pkl-pantry/blob/90b75bf0f48a604edaae92483ee8c3a2dfccaa5c/packages/org.openapis.v3/SecurityScheme.pkl#L24

https://github.com/apple/pkl-pantry/blob/90b75bf0f48a604edaae92483ee8c3a2dfccaa5c/packages/org.openapis.v3/SecurityScheme.pkl#L33-L44

This issue is to request that this could be fixed, so as to be able to produce valid OpenAPI schemas. (I would be glad to volunteer to make this fix if possible)

HT154:
This is definitely an issue. This repo is open for contribution, so don't hesitate to open a PR for a fix!

My reading of the spec is that both `name` and `in` are required if and only if `type == "apiKey"`. Should be fairly simple to encode that using type constraints.

nikeokoronkwo:
The pull request is live here: #91

PR #91, opened 2024-12-28
[org.openapis.v3] Security Scheme Update for HTTP/Bearer Auth

nikeokoronkwo:
This pull request addresses the issue of being able to make a correctly configured security scheme for HTTP security schemes like Bearer Auth Tokens.

For more information, see the original issue: #90

### DIFF
--- a/packages/org.openapis.v3/SecurityScheme.pkl
+++ b/packages/org.openapis.v3/SecurityScheme.pkl
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


packages/org.openapis.v3/SecurityScheme.pkl
```diff
@@ -30,10 +31,10 @@ type: SecuritySchemeType
 description: String?
 
 /// The name of the header, query or cookie parameter to be used.
-name: String
+name: String?((type == "apiKey").implies(this != null))
```
HT154, line 34: I think this condition may be reversed. As-is, it doesn't forbid these properties from being non-null when the type is not `apiKey`.

```suggestion
name: String?((this != null).implies(type == "apiKey"))
```

HT154:
Don't forget to bump the package patch version in PklProject. If there are local packages that depend on this one you will need to bump those versions as well and then run `./gradlew resolveProjects`.

packages/org.openapis.v3/SecurityScheme.pkl
```diff
@@ -30,10 +31,10 @@ type: SecuritySchemeType
 description: String?
 
 /// The name of the header, query or cookie parameter to be used.
-name: String
+name: String?((this != null).implies(type == "apiKey"))
 
 /// The location of the API key.
-`in`: "query"|"header"|"cookie"
+`in`: SecuritySchemeIn?((type == "apiKey").implies(this != null))
```
HT154, line 37: The type constrain here should match the one for `name`

HT154 (approved):
Looks good to me! This will need final approval and merge from the core team.

bioball (approved):
LGTM, thank you!

PR #98, opened 2025-02-15
Fix clashing class names when generating Pkl from OpenAPI/JSON Schema

bioball:
This fixes an issue where sometimes the code generators will generate multiple classes with the same name.

Also, the CRD generator API has changed slightly. In order to convert a listing element, the path should now end with `"[]"`.


packages/k8s.contrib.crd/PklProject
```diff
@@ -29,5 +29,5 @@ dependencies {
 }
 
 package {
-  version = "1.0.14"
+  version = "2.0.0"
```
bioball, line 32: The API of the converter changed (you need to use a different path sometimes when using a converters), so this is a major version bump.

CC @jackkleeman

packages/k8s.contrib.crd/tests/ModuleGenerator.pkl
```diff
@@ -28,11 +28,11 @@ local generator = (generate) {
...
     ["restateclusters.restate.dev"] {
-      [List("spec", "compute", "env", "env")] = EnvVar
+      [List("spec", "compute", "env", "[]")] = EnvVar
       [List("spec", "compute", "resources")] = ResourceRequirements
-      [List("spec", "security", "networkPeers", "ingress", "ingres")] = NetworkPolicy.NetworkPolicyPeer
-      [List("spec", "security", "networkPeers", "admin", "admin")] = NetworkPolicy.NetworkPolicyPeer
-      [List("spec", "security", "networkPeers", "metrics", "metric")] = NetworkPolicy.NetworkPolicyPeer
+      [List("spec", "security", "networkPeers", "ingress", "[]")] = NetworkPolicy.NetworkPolicyPeer
+      [List("spec", "security", "networkPeers", "admin", "[]")] = NetworkPolicy.NetworkPolicyPeer
+      [List("spec", "security", "networkPeers", "metrics", "[]")] = NetworkPolicy.NetworkPolicyPeer
```
bioball, line 35: I needed a way to disambiguate "this is the element of a listing" when resolving naming conflicts.

Previously, the code generator was just appending the same name twice, for "element of a listing", but that can be confused with a normal nested property.

Instead, this represents the listing element as `"[]"`.

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
```diff
@@ -215,22 +216,50 @@ local function determineTypeName(path: List<String>, candidateName: String, exis
...
+  else if (path.last == "[]")
+    path.dropLast(1).lastOrNull?.ifNonNull((it) -> utils.pascalCase(singularize.singularize(it))) ?? "Item"
+  else
+    utils.pascalCase(path.last)
+
+local function dropLast(path: List<String>) =
+  if (path.last == "[]")
+    path.dropLast(2)
+  else
+    path.dropLast(1)
```
stackoverflow, line 242: These functions seem to be a duplicate of `k8s/contrib/crd/internal/ModuleGenerator.pkl`.

stackoverflow (approved):

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
bioball: They are; we should ideally consolidate these two into one. But that's out of scope for this PR.

PR #105, opened 2025-03-20
[icalendar] Add package

HT154:
This PR introduces functionality for handling [RFC 5545 Recurrence Rules](https://datatracker.ietf.org/doc/html/rfc5545#autoid-42).
In general, attempts are made to conform strictly to the RFC where possible, but there are some small deviations.

Included functionality:
* Modules representing the `DATE`, `DATE-TIME`, `RECUR`, and `TIME` property value data types
* Rendering of each of these types to their serialized form via `toString()`
* Parsing of `Date`, `DateTime`, and `Time` objects from their string representations

Omitted functionality that can be added in the future:
* Parsing of `Recur` objects from their string representations
* Support for other structured property value data types
* Support for calendar objects or specific calendar components (eg. events, to-dos, alarms, etc.)

### DIFF
--- a/packages/icalendar/type/Recur.pkl
+++ b/packages/icalendar/type/Recur.pkl
@@ -1,0 +1,194 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+/// A recurrence rule.
+///
+/// https://datatracker.ietf.org/doc/html/rfc5545#section-3.3.10
+module icalendar.type.Recur
+
+import "Date.pkl"
+import "DateTime.pkl"
+
+typealias Frequency = "SECONDLY"|"MINUTELY"|"HOURLY"|"DAILY"|"WEEKLY"|"MONTHLY"|"YEARLY"
+
+typealias WeekdayString = "SU"|"MO"|"TU"|"WE"|"TH"|"FR"|"SA"
+
+class Weekday {
+  weekday: WeekdayString
+  weeknum: Int(this != 0)?
+
+  function toString(): String = List(
+    weeknum?.ifNonNull((n) -> if (n as Int > 0) "+" else ""),
+    weeknum?.toString(),
+    weekday
+  ).filterNonNull().join("")
+}
+
+typealias ByWeekday = WeekdayString|* Weekday
+
+dtstart: (
+  * DateTime(utc.implies(tzid == null))
+  |Date(tzid == null)
+  )?
+
+tzid: String?
+
+/// The FREQ rule part identifies the type of recurrence rule.
+freq: Frequency
+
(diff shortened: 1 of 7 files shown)


packages/icalendar/type/Recur.pkl
```diff
@@ -0,0 +1,198 @@
...
+dtstart: (
+  * DateTime(utc.implies(tzid == null))
+  |Date(tzid == null)
+  )?
+
+tzid: String?
+
+/// The FREQ rule part identifies the type of recurrence rule.
+freq: Frequency(
+    !(count != null && until != null) || throw("count and until may not both be set")
```
bioball, line 50: [nit] move this constraint to either `count` or `until`, because this doesn't have anything to do with `freq`

packages/icalendar/tests/Recur.pkl-expected.pcf
```diff
@@ -0,0 +1,34 @@
+examples {
+  ["with date"] {
+    """
+    DTSTART;VALUE=DATE:20000101
+    RRULE:FREQ=YEARLY;INTERVAL=1;UNTIL=20250101;BYDAY=MO,-1FR;BYMONTHDAY=15,-15;BYYEARDAY=100,-100;BYMONTH=1,11;WKST=MO;BYSETPOS=-1,1
+    """
+    "Expected value of type `* DateTime(utc.implies(tzid == null))   |Date(tzid == null)`, but got a different `icalendar.Date`. Value: new ModuleClass { y = ?; m = ?; d = ? }"
```
bioball, line 7: Error messages shouldn't be part of an API (if/when we improve error messages in pkl-core, we don't want our packages to break). Can you change these tests to something like:

```pkl
    module.catchOrNull(() -> new Recur {
      dtstart = DateTime.DateTime("20000101T000000")
      freq = "YEARLY"
      until = DateTime.DateTime("20250101T000000Z")
    }.toString()) != null
```

packages/icalendar/type/Recur.pkl
```diff
@@ -0,0 +1,198 @@
...
+    !isEmpty,
+    freq == "MONTHLY" || freq == "YEARLY",
+    !(freq == "YEARLY" && byweekno != null)
+  )?
+
+/// The BYMONTHDAY rule part specifies one or more days of the month.
+///
+/// Negative values represent days counting back from the end of a month.
+/// For example, -10 represents the tenth to the last day of the month.
+bymonthday: Listing<Int(relativeRange(31))>(
```
bioball, line 120: For each of these, I think it's a little better to split these into separate constraints, because Pkl gives a better error message when this fails.

Otherwise, you get a message like ``Type constraint `relativeRange(31)` failed. Value: 0``, and this doesn't make too much sense to users.

```suggestion
bymonthday: Listing<Int(isBetween(-31, 31), this != 0)>(
```

packages/icalendar/type/Recur.pkl
```diff
@@ -0,0 +1,198 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+/// A recurrence rule.
+///
+/// https://datatracker.ietf.org/doc/html/rfc5545#section-3.3.10
+module icalendar.Recur
```
bioball, line 19: This and everywhere else: move into the root package dir (should be packages/icalendar/Recur.pkl)

HT154: I was hoping to separate out the iCalendar data types (DATE, TIME, DATE-TIME, RECUR, etc.) from the (future) calendar object or components types (VCALENDAR, VEVENT, VTODO, etc.) by directory without having to cut a major release down the road.

bioball: Makes sense!

bioball (approved):
LGTM

PR #106, opened 2025-03-25
[icalendar] Fix byday constraints

HT154:
Always one little thing you notice right after something ships! RFCs will try to bite you when they think you're not paying attention.


bioball (approved):
LGTM

PR #109, opened 2025-05-14
Improve error message provided on conversion failure

thomaspurchas:
When a conversion failure occurs deepToTyped creates a path to the location of the failure in the object that's undergoing conversion. Now the error indicates both the type and location of the error making it much easier to debug problems.

Example of previous error:

```
    Expected "String" but got "Dynamic"
```

Example of new error:

```
    Expected "String" but got "Dynamic" at 
    .metadata.managedFields[0].fieldsV1
```

### DIFF
--- a/packages/pkl.experimental.deepToTyped/deepToTyped.pkl
+++ b/packages/pkl.experimental.deepToTyped/deepToTyped.pkl
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


packages/pkl.experimental.deepToTyped/deepToTyped.pkl
```diff
@@ -41,10 +41,26 @@ function attemptApply(type: Class|TypeAlias, value: Any): Any|ConversionFailure
 
 
 class ConversionFailure {
+  fixed Error = "\(message) at \n\(renderedPath)"
```
bioball, line 44:
```suggestion
  fixed error = "\(message) (at path \(renderedPath))"
```

packages/pkl.experimental.deepToTyped/deepToTyped.pkl
```diff
@@ -30,7 +30,7 @@ function apply(type: Class|TypeAlias, value: Any): Any =
   let (result =
     attemptApply(type, value)
   )
-    if (result is ConversionFailure) throw(result.message) else result
+    if (result is ConversionFailure) throw(result.Error) else result
```
bioball, line 33: Nit: properties should be camel cased

```suggestion
    if (result is ConversionFailure) throw(result.error) else result
```

bioball (approved):
LGTM, thanks! Good improvement.

PR #115, opened 2025-07-07
Add labels field to Prometheus rule groups

carlfriess:
Prometheus has added the labels field to alerting and recording rule groups and it's quite useful.
See https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#rule_group

### DIFF
--- a/packages/io.prometheus/PklProject
+++ b/packages/io.prometheus/PklProject
@@ -18,5 +18,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.3.0"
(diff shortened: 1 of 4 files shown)


HT154 (commented):
Don't forget to update the package's version in its PklProject file. This should be a minor version bump.

carlfriess:
> Don't forget to update the package's version in its PklProject file. This should be a minor version bump.

Done! :)

bioball (approved):
Thanks!

PR #116, opened 2025-07-08
[k8s.contrib.crd] Add generation of k8sSchema module

HT154:
This PR adds generation of a schema module similar to [the one from pkl-k8s](https://github.com/apple/pkl-k8s/blob/main/generated-package/k8sSchema.pkl), which enables easy import and use in [`k8s.contrib`'s converter module](https://pkl-lang.org/package-docs/pkg.pkl-lang.org/pkl-pantry/k8s.contrib/current/convert/index.html#resourcesToConvert)

### DIFF
--- a/packages/k8s.contrib.crd/generate.pkl
+++ b/packages/k8s.contrib.crd/generate.pkl
@@ -83,8 +83,14 @@ k8sVersion: String(semver.isValid(this)) = "1.0.1"
 /// ```
 k8sImportPath: String = read?("prop:k8sImportPath") ?? "package://pkg.pkl-lang.org/pkl-k8s/k8s@\(k8sVersion)#"
 
+/// The base API group.
+///
+/// Used to determine the output directory structure and which parts should be included in the path.
 baseApiGroup: String = read?("prop:baseApiGroup") ?? "k8s.io"
 local baseApiGroupModulePrefix = baseApiGroup.split(".").reverse().join(".")
+
+/// Enable or disable generating a k8sSchema.pkl at the root of the output.
+generateSchema: Boolean = read?("prop:generateSchema")?.toBoolean() ?? false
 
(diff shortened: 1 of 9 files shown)


packages/k8s.contrib.crd/internal/SchemaGenerator.pkl
```diff
@@ -0,0 +1,102 @@
...
+            new TypeNode.DeclaredTypeNode {
+              name { parts { new { value = "String" } } }
+            }
+            new TypeNode.DeclaredTypeNode {
+              name { parts { new { value = "Mapping" } } }
+              typeArguments {
+                new TypeNode.DeclaredTypeNode {
+                  name { parts { new { value = "String" } } }
+                }
+                new TypeNode.BuiltInTypeNode { type = "unknown" }
```
bioball, line 57: This should just be `K8sResource` now.

It used to matter (up until Pkl 0.26) because setting a declared type will execute each import when evaluating the map. But, with the lazy typecheck change, this is no longer a concern.

packages/k8s.contrib.crd/generate.pkl
```diff
@@ -148,11 +150,17 @@ fixed modules: Listing<ModuleGenerator> = new {
...
 
+fixed schema: SchemaGenerator = new {
+  crds = _crds
+  baseApiGroup = module.baseApiGroup
+}
+
 output {
   text = throw("The JSON Schema generator only works with multiple-file output. Try running again with the -m option.")
   files {
+    ["k8sSchema.pkl"] = schema.moduleNode.output
```
bioball, line 161: I think it'd be good to put this under a flag; many users likely would not care for this.

bioball (changes requested):
Mostly looks good! Just two comments

bioball (approved):
LGTM!

PR #120, opened 2025-09-03
[com.circleci.v2] Add converter for Pkl's Duration type to CircleCI duration settings format

yyuu:
In the schema of `com.circleci.v2`, it's declaring its configuration parameter of `no_output_timeout` as Pkl's [Duration](https://pkl-lang.org/package-docs/pkl/current/base/Duration.html) type. However, given type was not configured in `converters` configuration, therefore Pkl's YAML renderer couldn't render the value.

According to CircleCI documentation, `no_output_timeout` is defined as follows. I've added some some conversion rule by making `min` as only special case.
> Elapsed time the command can run without output. The string is a decimal with unit suffix, such as "20m", "1.25h", "5s".

refs:
* https://circleci.com/docs/reference/configuration-reference/

### DIFF
--- a/packages/com.circleci.v2/PklProject
+++ b/packages/com.circleci.v2/PklProject
@@ -17,5 +17,5 @@
 amends "../basePklProject.pkl"
 
 package {
-  version = "1.6.0"
+  version = "1.6.1"
 }


packages/com.circleci.v2/Config.pkl
```diff
@@ -813,6 +813,8 @@ output {
   renderer = new YamlRenderer {
     converters {
       [AbstractStep] = (it) -> Map(it.__name__, it.toMap())
+      [Duration] = (it) ->
+        let (unit = if (it.unit == "min") "m" else it.unit) "\(it.value)\(unit)"
```
bioball, line 817: Hm, not sure if CircleCI accepts days, nanoseconds, microseconds or milliseconds. Let's coerce these to units that are described by the documentation.

```suggestion
        let (dur =
          if (it.unit is "d") it.toUnit("min")
          else if (it.unit is "us"|"ms"|"ns") it.toUnit("s")
          else it
        )
        let (unit = if (dur.unit == "min") "m" else dur.unit)
          "\(dur.value)\(unit)"
```

bioball (changes requested):
Thanks, good fix! Just one minor comment.

Also, can you bump the package version?

PR #121, opened 2025-10-16
[com.circleci.v2] Add m4pro resource classes

HT154:


packages/com.circleci.v2/Config.pkl
```diff
@@ -218,6 +220,7 @@ typealias ResourceClass =
   |"gpu.nvidia.medium"
   |"gpu.nvidia.large"
   |"windows.gpu.nvidia.medium"
+  |String
```
HT154: Added `String` here to avoid needing to update the package to match CircleCI changes in the future.

PR #123, opened 2025-11-06
[pkl.tar] Add package

HT154:


packages/pkl.tar/File.pkl
```diff
@@ -0,0 +1,407 @@
...
+
+local const ustarMaxLinkLength: UInt = 100
+local const ustarMaxUidGid: UInt = 2 ** 21 as UInt
+local const ustarMaxSizeTimestamp: UInt = 2 ** 33 as UInt
+
+/// File type marker, also known as `linkflag`.
+fixed type: Char
+
+/// The numeric byte value of [type].
+fixed typeByte: UInt8 = type.encodeToBytes("UTF-8")[0]
```
bioball:
```suggestion
fixed typeByte: UInt8 = type.codePoints[0]
```

packages/pkl.tar/File.pkl
```diff
@@ -0,0 +1,407 @@
...
+      ["uname"] = info.uname
+    }
+    when (!(info.gname is PosixPortableString)) {
+      ["gname"] = info.gname
+    }
+  }
+}
+
+/// The payload bytes of this file, padded to the nearest 512 bytes.
+fixed paddedBytes: Buffer(length % blockLength == 0) = List()
```
bioball:
```suggestion
abstract fixed paddedBytes: Buffer(length % blockLength == 0)
```

packages/pkl.tar/File.pkl
```diff
@@ -0,0 +1,407 @@
...
+fixed type: Char
+
+/// The numeric byte value of [type].
+fixed typeByte: UInt8 = type.encodeToBytes("UTF-8")[0]
+
+/// File information.
+info: Info?
+
+/// Pax archive extended header data fields.
+paxFields: PaxFields = new {
```
bioball: We advise against having object property types with default values.

Can we do this instead?

```pkl
paxFields: PaxFields

fixed effectivePaxFields: PaxFields = (paxFields) {
  when (info != null) {
    when (info.gid > ustarMaxUidGid) {
      ["gid"] = info.gid.toString()
    }
    when (info.uid > ustarMaxUidGid) {
      ["uid"] = info.uid.toString()
    }
    when (info.mtime != info.mtime.truncate()) {
      ["mtime"] = formatPaxNumber(info.mtime)
    }
    when (!(info.uname is PosixPortableString)) {
      ["uname"] = info.uname
    }
    when (!(info.gname is PosixPortableString)) {
      ["gname"] = info.gname
    }
  }
}
```

packages/pkl.tar/File.pkl
```diff
@@ -0,0 +1,407 @@
...
+  )
+    if (paxHeader.paxFields.isEmpty)
+      List()
+    else
+      checksum(paxHeader.makeHeader(path, pax)) + paxHeader.paddedBytes
+
+/// File metdata information.
+class Info {
+  /// File permissions mode.
+  mode: Set<Mode>
```
bioball: You can also use octal literals in Pkl, so, you could just leave this as type `Int(isBetween(0o000, 0o777))`; and users would just do `0o444` etc.

packages/pkl.tar/examples/test_examples.sh
```diff
@@ -0,0 +1,14 @@
...
+
+cd "$(dirname "$0")"
+
+for file in archive.*.pkl; do
+  echo "+ pkl eval $file | tar tv"
+  pkl eval $file | tar tv -v
+  echo
+done
+
+echo "PASS"
```
bioball: Good test! Let's add a one-off Gradle task to run this:

```kt
val testTarExamples by tasks.registering(Exec::class) {
  executable = file("packages/pkl.tar/examples/test_examples.sh").absolutePath

  inputs.files(fileTree("packages/pkl.tar/"))

  val outputFile = layout.buildDirectory.file("testTarExample.txt")
  outputs.file(outputFile)

  doLast {
    outputFile.get().asFile.writeText("OK")
  }
}

tasks.test {
  dependsOn(testTarExamples)
}
```

packages/pkl.tar/examples/archive.pax.pkl
```diff
@@ -0,0 +1,52 @@
...
+    target = "\(superLongPathParent)/regularFile.txt"
+  }
+  ["\(superLongPathPrefix)/someDir/otherDir"] = new File.Directory {}
+}
+
+renderer {
+  defaultFileInfo {
+    uid = 2 ** 21 as UInt
+    gid = 2 ** 21 as UInt
+    mtime = 1767225600000.123456 // 2026-01-01T00:00:00.123456Z
```
bioball: Floating point math makes me nervous.

Maybe better to just have

```pkl
  /// File modification time in epoch nanoseconds.
  ///
  /// Truncated to whole seconds for UStar archives.
  /// Must be < 2**33 for UStar archives.
  mtime: Int
```

bioball (changes requested):
Fantastic! This looks really awesome.

I did a first initial pass and have some comments.

packages/pkl.tar/File.pkl
HT154: Hmm. For several of the subclasses (`Link`, `SymbolicLink`, and `Directory` specifically), this would just return an empty list, which would mean this gets repeated a bunch.

packages/pkl.tar/examples/archive.pax.pkl
HT154: Pax archives support encoding timestamps with fractional seconds and if we restricted this to integers it would be the only part of the pax standard we're explicitly not supporting.

Thoughts on this instead?
```pkl
mtime: UInt|String(matches(Regex(#"\d+(\.\d+)?"#)))
```

bioball: That sounds good to me.

By the way, epoch nanos would still qualify as fractional seconds (with nanosecond precision)

packages/pkl.tar/File.pkl
```diff
@@ -0,0 +1,401 @@
...
+  /// * 0o0010 - Executable by group
+  /// * 0o0004 - Readable by other
+  /// * 0o0002 - Writable by other
+  /// * 0o0001 - Executable by other
+  mode: Int(isBetween(0o0000, 0o7777))
+
+  /// File modification time in epoch seconds.
+  ///
+  /// Truncated to whole seconds for UStar archives.
+  /// Must be < 2**33 for UStar archives.
```
bioball: This should be updated

HT154: String is actually better here either way since `Number.toFixed()` doesn't support enough digits to epoch seconds with nanosecond precision.

packages/pkl.tar/examples/test_examples.sh
HT154: What's the best way to handle no `pkl` executable being present in CI? Should we just install one? Or is there a way to set an env var to ask the script to exec `jpkl` out of the class path somehow?

packages/pkl.tar/tar.pkl
```diff
@@ -0,0 +1,141 @@
...
+              [pax.globalexthdrName] = new File.GlobalExtendedHeader {
+                paxFields = pax.globalHeaderFields
+              }
+              ...value
+            }
+      )
+        fs
+          .fold(List(), (result, path, file) ->
+            let (tarFile = toTar(file, defaultFileInfo))
+              (result |> File.maybeAppend(pax != null, () -> tarFile.makePax(path, pax!!)))
```
srueg: I'm getting an "out of bounds" array access here:
```
java.lang.ArrayIndexOutOfBoundsException%3A%20Index%20-17%20out%20of%20bounds%20for%20length%2019

–– Pkl Error ––
Index -17 out of bounds for length 19

105 | (result |> File.maybeAppend(pax != null, () -> tarFile.makePax(path, pax!!)))
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.tar.tar#Renderer.renderValue.<function#3> (/Users/srueg/Development/work/iCloud/icloud-monorepo/pkl/tar/tar.pkl:105)

104 | let (tarFile = toTar(file, defaultFileInfo))
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.tar.tar#Renderer.renderValue.<function#2> (/Users/srueg/Development/work/iCloud/icloud-monorepo/pkl/tar/tar.pkl:104)

102 | fs
      ^^
at pkl.tar.tar#Renderer.renderValue.<function#1> (/Users/srueg/Development/work/iCloud/icloud-monorepo/pkl/tar/tar.pkl:102)

90 | let (
     ^^^^^
at pkl.tar.tar#Renderer.renderValue (/Users/srueg/Development/work/iCloud/icloud-monorepo/pkl/tar/tar.pkl:90)

116 | function renderDocument(value: Any): Bytes = renderValue(value) + File.endOfArchive.toBytes()
                                                   ^^^^^^^^^^^^^^^^^^
at pkl.tar.tar#Renderer.renderDocument (/Users/srueg/Development/work/iCloud/icloud-monorepo/pkl/tar/tar.pkl:116)

271 | if (renderer is BytesRenderer) renderer.renderDocument(value) else text.encodeToBytes("UTF-8")
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.base#FileOutput.bytes (https://github.com/apple/pkl/blob/0.30.0/stdlib/base.pkl#L271)

Pkl 0.30.0 (macOS 26.2, Java 21.0.9)

java.lang.ArrayIndexOutOfBoundsException: Index -17 out of bounds for length 19
```

HT154: This is resolved by https://github.com/apple/pkl/pull/1336. We're planning on releasing Pkl 0.30.1 once this merges.

packages/pkl.tar/File.pkl
```diff
@@ -0,0 +1,404 @@
...
+      // suffix is the longest suffix under 100 bytes that starts after a /
+      let (suffix = name.takeLast(100).dropWhile((char) -> char != 0x2f).drop(1))
+      // prefix is the rest of the string not including the joining /
+      let (fullPrefix = name.dropLast(suffix.length + 1))
+      let (prefix = fullPrefix.takeLast(155))
+        if (fullPrefix.length > 155)
+          if (pax == null)
+            throw("Path `\(path)` too long for UStar archive")
+          else
+            null
```
srueg: This is causing the following error:

```
Cannot find property `first` in object of type `Null`.

89 | pathParts.first, // name [0, 100)
     ^^^^^^^^^^^^^^^
at pkl.tar.File#makeBaseHeader.<function#1> (file:///<redacted>/pkl/tar/File.pkl, line 89)
```

PR #129, opened 2025-11-19
[org.json_schema.contrib] Handle Null typed fields

thomasquinn:
Currently, a `new JsonSchema { type = "null" }` generates Pkl code matching `Any`. This change explicitly handles `type == "null"` instead of allowing it to fallthrough to the `Any` case.


thomasquinn:
Updated the k8s.contrib.crd dependency on this package as well, so `./gradlew build` should pass now (or at least it does locally at least), apologies for not testing that earlier.

PR #130, opened 2025-11-19
Add GitHub Actions

bioball:
This PR introduces four new packages that revolve around using GitHub Actions.

In particular, it forks https://github.com/stefma/pkl-gha, but with some refactorings. Thanks for the ground work here, @StefMa!

The new packages:

* com.github.actions -- templates for defining workflows, jobs, environments, action metadata, and also hosts the core set of actions maintained by GitHub
* com.github.actions.community -- community-defined actions defined in Pkl (this package is entirely code-generated)
* com.github.actions.contrib -- tools for using GitHub Actions, in particular, a code generator for typesafe actions
* io.github.typesafegithub -- Pkl types for typesafegithub

## Differences between StefMa/pkl-gha

There have been some things changed from the original pkl-gha repo. For those migrating, those are:

1. `Workflow#DefaultJob` -> moved to own module `Job.pkl`
1. `Workflow#AbstractStep` -> moved to own module `AbstractStep.pkl`
1. `Workflow#Step` -> moved to own module `Step.pkl`
1. `Workflow#TypedStep` -> moved to own module `AbstractTypedStep.pkl`
1. `Workflow#ReusableJob` -> removed, users should just use `Job.pkl`
1. `Context.pkl` -> renamed to `context.pkl`
1. `Actions.pkl` -> renamed to `catalog.pkl`, property names changed to full action name and version.
1. All actions within `actions/` moved into their own module, and the path has been adjusted to include the repo name and version. Also, they are code-generated!
1. More actions have been added to the `actions/`

### DIFF
--- a/packages/com.github.actions/Action.pkl
+++ b/packages/com.github.actions/Action.pkl
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/com.github.actions/actions/cache/restore/v4/CacheRestore.pkl
+++ b/packages/com.github.actions/actions/cache/restore/v4/CacheRestore.pkl
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/com.github.actions/actions/cache/save/v4/CacheSave.pkl
+++ b/packages/com.github.actions/actions/cache/save/v4/CacheSave.pkl
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/com.github.actions/actions/cache/v4/Cache.pkl
+++ b/packages/com.github.actions/actions/cache/v4/Cache.pkl
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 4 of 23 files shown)


packages/com.github.actions.contrib/PklProject
```diff
@@ -0,0 +1,30 @@
...
+  ["com.github.actions"] = import("../com.github.actions/PklProject")
+  ["io.github.typesafegithub"] = import("../io.github.typesafegithub/PklProject")
+  ["deepToTyped"] = import("../pkl.experimental.deepToTyped/PklProject")
+  ["syntax"] = import("../pkl.experimental.syntax/PklProject")
+}
+
+package {
+  version = "1.0.0"
+  exclude {
+    "scripts/"
```
HT154: The scripts dir doesn't exist in this package

HT154 (approved):
Looks great! One comment, but approving to unblock

StefMa:
> I get that they're different use-cases, but, they're both related to GitHub Actions. Is there a specific reason we don't want them in the same package?

Because it complicates things for especially startes.
The "product" is named *GitHub Actions*. 
You can create *Actions*.
Those can then be used within a *Worfklow* as a *Step*.

This is confusing IMHO.
GitHub Actions "users" mainly want to create Workflows. 
99,99% of `pkl-gha` users are not interested in creating Actions directly.
They already have to learn a new tool (pkl).
So I think we would just make it more complicated by proving more than necceary for those users.

> Ah, are some of these old versions not useable anymore? Good idea, though, probably makes sense to just roll out with the latest version of each of these.

When you write an Action (in JavaScript, wich is the main case), you define which `node` version you use.
For the default runners, `node20` and `node24` is support AFAIK.
But even node20 will get deprecated in March 2026.
See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
I just assume that old versions using an old version of node that are not installed on the runners and therefore fail.


> Sounds good, but, also not sure how feasible this is? Given how often new versions can be published.

Yeah, this is the problem :( 
That is why I also went with only major releases for the first implementation.
Just let us keep that in mind for now. We might fnid a solution later.
Just gonna drop those links here:
https://unit42.paloaltonetworks.com/github-actions-supply-chain-attack/
https://github.com/actions/publish-immutable-action

> This means that you need to import at least two packages for the most basic routines.
> That feels bad to me?
> We can always mark these things as deprecated.

Can't we add the "action" package as a dependency to the `github.action` package?
Then the `gitub.action` package contains still the `catalog` that "just delegates" to all the actions within the action package?
Would this work? 🤔 
Then we have a clean separation:
com.github.action: Contains just the Workflow stuff
com.github.action.Actions: Contains all the actions
But since com.github.action depend on `*.Actions`, all of them are available via the catalog... 🤔 
Not sure if this works without adding two dependencies 🤔 But I guess so..?! 

> It sounds pretty cool to generate Pkl packages on demand!

Its just a prototyp I hacked quickly together, but visit e.g.:
https://pkl-gha-pkgs.vercel.app/actions/setup-go@5.0.0
While `actions/setup-go` can be any repo and the `version` can be any semver version.
https://pkl-gha-pkgs.vercel.app/actions/setup-go@5.0.0.zip
will of course work as well.

So I've already something in place if we want/need this.
Currently using a pretty naiv implementation of generating the Actions.pkl file, but of course it can be changed to your generation package as soon as published.
I can share the code (Go) with you if you want. 
Maybe you can deploy something on `gh-pkg.pkl-lang.org/` 🤩 

---

Over the weekend I had another thing in mind:
The `with` block doesn't "scale" yet.
Currently users are "stuck" with the options we have.
Since we are using just the major version, it might happen that a minor introduce a new option.
If a user wants to use this option, he has to wait until its published.
To "fix" this, we might can add something like `additionalWith: Mapping<String, Boolean...` which will merged to the `with`... 🤔 Then we are more flexible at least in that regards...

PR #139, opened 2025-12-17
Clean up Gradle usage

StefMa:
* Remove buildSrc
* Move all plugins to plugins in version catalog
---

The `buildSrc` has a down side, that it has to be compiled before the actual project get compiled.
This "slows down" the project. Besides its not a best practise, I guess.

So I removed it and inlined the diffplug class directly into to build script.
This speed up things a bit.

I have not idea why I could remove the Formatter dependency.
I've the feeling it is somehow part of the plugin because:
`plugin` depends on
`tools` depends on 
`pkl-cli` depends on
`pkl-formatter` 

So it seems it is part of it?! 🤷 
I can also add it back via the buildscript block if wanted.

I also updated the `plugins` block by using `aliases`.


bioball (approved):
Nice, thanks!

PR #140, opened 2025-12-17
Remove prepareCiGit task

StefMa:
It's not used anymore. Even if you want to publish as pkl-ciCd again, it should be managed within the workflows, not as a gradle task IMHO...


bioball (approved):
LGTM, thanks!

PR #166, opened 2026-03-19
Add Cilium k8s CRDs

bmurray:
Cilium is a widely used CNI for k8s clusters. We're trying to keep all our k8s configs in pkl, so we needed the pkl schemas. I used the k8s.contrib.crd generator to generate the CRDs. I also included some tests, similar to the k8s.networking.gateway CRDs.

### DIFF
--- a/packages/io.cilium/generate.sh
+++ b/packages/io.cilium/generate.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+set -euo pipefail
+
+BASE_URL="https://raw.githubusercontent.com/cilium/cilium/refs/heads/main/pkg/k8s/apis/cilium.io/client/crds"
+GENERATE_PKG="package://pkg.pkl-lang.org/pkl-pantry/k8s.contrib.crd@4.2.1#/generate.pkl"
+
+V2_CRDS=(
+  ciliumbgpadvertisements.yaml
+  ciliumbgpclusterconfigs.yaml
+  ciliumbgpnodeconfigoverrides.yaml
+  ciliumbgpnodeconfigs.yaml
+  ciliumbgppeerconfigs.yaml
+  ciliumcidrgroups.yaml
+  ciliumclusterwideenvoyconfigs.yaml
+  ciliumclusterwidenetworkpolicies.yaml
+  ciliumegressgatewaypolicies.yaml
+  ciliumendpoints.yaml
+  ciliumenvoyconfigs.yaml
+  ciliumidentities.yaml
+  ciliumloadbalancerippools.yaml
+  ciliumlocalredirectpolicies.yaml
+  ciliumnetworkpolicies.yaml
+  ciliumnodeconfigs.yaml
+  ciliumnodes.yaml
+)
+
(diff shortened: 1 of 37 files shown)


packages/io.cilium/generate.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -euo pipefail
+
+BASE_URL="https://raw.githubusercontent.com/cilium/cilium/refs/heads/main/pkg/k8s/apis/cilium.io/client/crds"
```
HT154, line 4: This script _must_ be run from the package root.

```suggestion
set -euo pipefail

SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
cd "$SCRIPT_DIR"

BASE_URL="https://raw.githubusercontent.com/cilium/cilium/refs/heads/main/pkg/k8s/apis/cilium.io/client/crds"
```

packages/io.cilium/generate.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "Generating v2/$crd..."
+  pkl eval "$GENERATE_PKG" -m . -p source="$BASE_URL/v2/$crd" -p baseApiGroup="cilium.io"
+done
+
+for crd in "${V2ALPHA1_CRDS[@]}"; do
+  echo "Generating v2alpha1/$crd..."
+  pkl eval "$GENERATE_PKG" -m . -p source="$BASE_URL/v2alpha1/$crd" -p baseApiGroup="cilium.io"
+done
+
+echo "Done!"
```
HT154, line 44: Much nicer to format inline when generating code!

```suggestion
done

echo "Formatting..."
cd ../..
./gradlew spotlessApply

echo "Done!"
```

packages/io.cilium/v2/CiliumBGPAdvertisement.pkl
```diff
@@ -0,0 +1,153 @@
...
+
+  /// operator represents a key's relationship to a set of values. Valid operators are In, NotIn, Exists
+  /// and DoesNotExist.
+  operator: "In"|"NotIn"|"Exists"|"DoesNotExist"
+
+  /// values is an array of string values. If the operator is In or NotIn, the values array must be
+  /// non-empty. If the operator is Exists or DoesNotExist, the values array must be empty. This array is
+  /// replaced during a strategic merge patch.
+  values: Listing<String>?
+}
```
HT154, line 141: These classes (and others like `NodeSelector`) are repeated several times and are (largely) identical. It would be best to pull these out into a common file so that a `Selector` or `MatchExpression` from a `CiliumBGPAdvertisement` can be reused for `CiliumEgressGatewayPolicy` and others.

HT154 (changes requested):
Apologies for the delay on this. We needed to discuss how we intend to support CRDs like this in pkl-pantry. Here's how we're approaching this:
* We will happily accept generated packages containing K8s CRD.
* We (the Pkl team) will rely on community contributions to keep these packages up to date with their upstream software.

PR #169, opened 2026-04-07
[pkl.github.dependabotManagedActions] Allow customization of Dependabot settings for GHA

### LINKED ISSUE #167: [pkl.github.dependabotManagedActions] Action version updates in generated `dependabot.yml` should be configurable
Right know its hardcoded to updated or minor and patch versions:
https://github.com/apple/pkl-pantry/blob/d37f1d5cba2b4f0f4facde54e2e80d0c0e74c011/packages/pkl.github.dependabotManagedActions/DependabotManagedActions.pkl#L115-L122

Not everyone wants that.
We might want this for stabiliy reasons, but I guess it would make sense to make it configurable.
I'm not sure how to do this nicely. Any recommendation for a nice pkl API syntax is appreciated.

HT154:
Resolves #167

### DIFF
--- a/packages/pkl.github.dependabotManagedActions/DependabotManagedActions.pkl
+++ b/packages/pkl.github.dependabotManagedActions/DependabotManagedActions.pkl
@@ -101,7 +101,8 @@ dependabot: Dependabot
 /// Dependabot configuration for the `github-actions` ecosystem.
 dependabotGitHubActions: Dependabot.Update(
   `package-ecosystem` == "github-actions",
-  directory == "/"
+  directory == "/",
+  ignore.contains(semverMajorIgnoreAll)
 ) = new {
   `package-ecosystem` = "github-actions"
   directory = "/"
@@ -109,12 +110,17 @@ dependabotGitHubActions: Dependabot.Update(
     interval = "weekly"
   }
   ignore {
-    new {
-      `dependency-name` = "*"
-      `update-types` {
-        "version-update:semver-major"
-      }
-    }
+    semverMajorIgnoreAll
+  }
+  cooldown {
+    `default-days` = 7
+  }
+}
+
+hidden const semverMajorIgnoreAll: Dependabot.Ignore = new {
+  `dependency-name` = "*"
+  `update-types` {
+    "version-update:semver-major"
   }
 }
 
--- a/packages/pkl.github.dependabotManagedActions/DependabotManagedActions.pkl
+++ b/packages/pkl.github.dependabotManagedActions/DependabotManagedActions.pkl
@@ -95,8 +95,28 @@ workflows: Mapping<String, Workflow>
 
 /// Dependabot configuration.
 ///
-/// The `github-actions` ecosystem will be automatically added to this.
+/// [dependabotGitHubActions] will be automatically added to this.
 dependabot: Dependabot
+
+/// Dependabot configuration for the `github-actions` ecosystem.
+dependabotGitHubActions: Dependabot.Update(
+  `package-ecosystem` == "github-actions",
+  directory == "/"
+) = new {
+  `package-ecosystem` = "github-actions"
+  directory = "/"
+  schedule {
+    interval = "weekly"
+  }
+  ignore {
+    new {
+      `dependency-name` = "*"
+      `update-types` {
+        "version-update:semver-major"
+      }
+    }
+  }
+}
 
 local generator: LockfileGenerator = new {
   workflows = module.workflows.toMap().values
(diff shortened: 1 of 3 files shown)


StefMa (commented):
Hmm.. I am not quite sure with the API design 🤔
 
First a user can override the dependabot property, but suddenly the default github-action update config will be "injected". 
However, they can only update it if they override the github-action update config. But this requires to override a new config... Not mentioning that it's not even possible to remove this config at all (not sure if even desired? 🤔 🤷‍♀️). 

I've the feeling we should rather go into the direction of providing the default github-action update config as a property a user can use in case he overrides the dependabot config at all! 

So like "if dependabot config was override, use it as it is! No changes to it. But since we provide the github-action update config as a property, a user who override the dependabot config can simply inject it"... 🤔

What do you think?

bioball:
I think most of the settings for the github-actions ecosystem doesn't really need to be configurable:

* directory must be `/` (see https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/keeping-your-actions-up-to-date-with-dependabot#example-dependabotyml-file-for-github-actions)
* shouldn't update to a new major version, because that's supposed to be managed by Pkl

Maybe the only setting that makes sense to expose here is the interval if weekly is insufficient for some folks?

HT154:
I think it makes sense to be able to set a broad variety of options for the ecosystem. In fact, I think `package-ecosystem` and `directory` are the only ones that must be fixed. It doesn't make sense to allow major version updates, but why prevent users from setting additional or more specific `ignores` here?

bioball:
True, it might be meaningful also define another `ignores`. Although I don't know if we need to provide the entire `Update` as an API? Maybe we can just have the fields that are meaningful to configure as their own properties? Something like:

```pkl
dependabotGhaInterval: ScheduleInterval?

dependabotIgnores: Listing<Ignore>
```

I'm also not sure how needed this is (is this something you need @StefMa?)

StefMa:
> Although I don't know if we need to provide the entire Update as an API? Maybe we can just have the fields that are meaningful to configure as their own properties? 

I guess this is the right choice ☝️ 
Right know we are using (for gradle updates) the following config:
```pkl
      `package-ecosystem` = "gradle"
      directory = "/"
      schedule {
        interval = "cron"
        cronjob = "0 2 * * TUE"
      }
      registries = "*"
      `open-pull-requests-limit` = 50
      cooldown {
        `default-days` = 13
        include {
          "*"
        }
      }
```

In the past we had "the same" config for actions.
The `schedule`, the `open-pull-request-limit` (afaik its 5 by defailt), and the `cooldown` property.
`ignoring` is "new" for us because of using pkl now. However, I also think that make sense to expose.

Hmm.. looking at this, maybe it make sense to expose the whole `Update` as an API 😅 because setting up each property individually might look harder? 
Esepcially because you might have an dependabot override **anyways** (to add your "project" config like gradle, npm, etc..)? 🤔 

Just for example:
```pkl
dependabot = new {
  updates {
    new {
      `package-ecosystem` = "gradle"
      directory = "/"
      schedule {
        interval = "cron"
        cronjob = "0 2 * * TUE"
      }
      registries = "*"
      `open-pull-requests-limit` = 50
      cooldown {
        `default-days` = 13
        include {
          "*"
        }
      }
    }
  }
  registries {
    ["maven-google"] {
      type = "maven-repository"
      url = "https://dl.google.com/dl/android/maven2/"
    }
    ["gradle-plugin-portal"] {
      type = "maven-repository"
      url = "https://plugins.gradle.org/m2"
    }
  }
}
```
Our fuill config ☝️ 

Imagine, we need to setup the gituhb-action part seprately.
I guess this feels... wrong?!

HT154:
@StefMa @bioball I've revised this to use this pattern:
```pkl
dependabot: Dependabot?
const defaultDependabot: Dependabot = new { ... }
```

If `dependabot` is not set then `defaultDependabot`, but if users wish to supply their own config they can do so (optionally based on `defaultDependabot`)

StefMa (approved):
Nice! I like it this way!

bioball:
This feels even easier to get wrong? The purpose of this module is to set up your GitHub Actions such that they are also updated by dependabot; it shouldn't be possible to omit the github-actions ecosystem.

I actually think the original design works better, where we have something like:

```pkl
fixed semverMajorIgnore: Ignore = new {
  `dependency-name` = "*"
  `update-types` {
    "version-update:semver-major"
  }
}

githubActionsUpdate: Update(
  `package-ecosystem` == "github-actions",
  directory == "/",
  ignore.contains(semverMajorIgnore)
)
```

And then this is inserted into the effective dependabot config.

HT154:
@bioball Updated/reverted with the extra `ignore` validation. I also added a cooldown (with `default-days: 7`) since including one is best practice.

packages/pkl.github.dependabotManagedActions/DependabotManagedActions.pkl
```diff
@@ -95,9 +95,40 @@ workflows: Mapping<String, Workflow>
...
+  schedule {
+    interval = "weekly"
+  }
+  ignore {
+    new {
+      `dependency-name` = "*"
+      `update-types` {
+        "version-update:semver-major"
+      }
+    }
```
bioball:
```suggestion
    semverIgnoreAll
```

bioball (approved):

PR #171, opened 2026-04-11
[com.github.actions] Add more GitHub Actions

bioball:
Add the following actions:

* actions/cache/restore@v5
* actions/cache/save@v5
* actions/cache@v5
* actions/configure-pages@v6
* actions/dependency-review-action@v4

Also, update existing actions with new input properties, and also fix the github action generator to also look for `action.yml` at a branch.


packages/com.github.actions.contrib/ActionGenerator.pkl
```diff
@@ -105,11 +105,20 @@ local actionYamlUrl =
   else
     "https://raw.githubusercontent.com/\(owner)/\(repo)/refs/tags/\(version)/\(subpath)/action.yml"
 
+local actionYamlUrl2 =
```
StefMa, line 108: Maybe renaming them like tagBasedActiomYamlUrl and branchBasedActiomYamlUrl? 🤔

packages/com.github.actions/actions/cache/restore/v5/CacheRestore.pkl
```diff
@@ -0,0 +1,61 @@
...
+
+fixed action: "actions/cache/restore"
+
+fixed version: "v5"
+
+with: With?
+
+class With {
+  /// A list of files, directories, and wildcard patterns to restore
+  hidden pathList: Listing<String>
```
StefMa, line 29: Why is this hidden? Does this make any sense? 🤔 (There are others hidden too. I am not going to mention all)

scripts/generate-github-actions.pkl
```diff
@@ -100,10 +100,9 @@ local coreActions: Listing<String> = new {
   "actions/cache"
   "actions/cache/save"
   "actions/cache/restore"
-  "actions/configure-pages"
-  "actions/create-github-app-token"
+  //   "actions/create-github-app-token"
   "actions/create-release"
-  "actions/deploy-pages"
+  //   "actions/deploy-pages"
```
StefMa, line 105: Why is this commented out? 🤔

StefMa (commented):
Just a few questions 🙃.

packages/com.github.actions/actions/cache/restore/v5/CacheRestore.pkl
HT154: This is hidden because it shouldn't be in the rendered YAML. This is convenience for setting `path` which is a newline-delimited string. In cases where the typing indicates a delimited list property, the generator also produces the `fooList` property. Either the original `foo` or `fooList` may be set.

scripts/generate-github-actions.pkl
bioball: We might uncomment this in the future once these action types are added to the typesafegithub catalog!

This is an internal script; it's okay to be pragmatic here.

packages/com.github.actions.contrib/ActionGenerator.pkl
HT154: Agreed, but not a blocker for me

packages/com.github.actions/actions/checkout/v6/Checkout.pkl
```diff
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © 2025-2026 Apple Inc. and the Pkl project authors. All rights reserved.
+// Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
HT154: This seems odd to me. There are a few identical changes in other files too.

HT154 (approved):
LGTM overall, but there's something odd happening with the license headers

packages/com.github.actions/actions/checkout/v6/Checkout.pkl
bioball: This is an unfortunate consequence of code generation; the newly generated file blows away the old one, and spotless then adds a branch new license header.

bioball: Updated these files with sensible year ranges

PR #175, opened 2026-04-21
[com.github.actions] Add typed actions for CodeQL

bioball:
Also, fixes a bug where an input's description should be nullable.

Open question: do we want `github/codeql-action` inside the core catalog? Since these are maintained by GitHub, I feel like this makes sense.


HT154 (approved):
LGTM. I think it's reasonable to include these here.

PR #176, opened 2026-04-22
[com.github.actions] Add schema for codeql config

bioball:
Add a way to set codeql configuration directly in the `github/codeql-action/init` action.


packages/com.github.actions/github/codeql-action/init/Config.pkl
```diff
@@ -0,0 +1,114 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+/// Configuration values when initializing CodeQL.
+///
+/// See <https://docs.github.com/en/code-security/reference/code-scanning/workflow-configuration-options#custom-configuration-files>
+module com.github.actions.github.`codeql-action`.init.Config
```
HT154: Should this really be in the actions package? It feels like it warrants its own package entirely, similar to dependabot.

bioball: I initially started that way, but, as far as I can tell, the codeql config schema is used specifically by codeql-action: https://github.com/github/codeql-action/blob/1dcdb940d513da9c73c6de817f8f40c75934d14d/src/config/db-config.ts#L34

And I don't see it in the CodeQL website: https://codeql.github.com/docs/

Not sure if this is part of CodeQL itself, or just specific to the action. If it's the former, we can deprecate this and move it to its own package.

HT154 (approved):

PR #177, opened 2026-04-22
[com.github.dependabot] Refine Dependabot schema

bioball:


packages/com.github.dependabot/v2/Dependabot.pkl
```diff
@@ -128,7 +130,11 @@ class Update {
 
   /// Array of dependency patterns to include in a multi-ecosystem group. Required when using
   /// multi-ecosystem-group. Use '*' to include all dependencies.
-  patterns: Listing<String(!isEmpty)>(!isEmpty, isDistinct)?
+  ///
+  /// This field is required if [`multi-ecosystem-group`] is non-null.
```
HT154, line 134: The doc comment is not a bit redundant. It also does not indicate whether `patterns` is ever valid when not part of a group.

If it is valid when not in a group, it may be clearer to instead express this as:
```
`multi-ecosystem-group`: String(!isEmpty, patterns != null)?
```

packages/com.github.dependabot/v2/Dependabot.pkl
```diff
@@ -114,7 +114,9 @@ class Update {
   registries: (*Listing<String(!isEmpty)>(!isEmpty, isDistinct) | "*")?
 
   /// Schedule preferences
-  schedule: Schedule
+  ///
+  /// Required if [`multi-ecosystem-group`] is null.
+  schedule: Schedule?((`multi-ecosystem-group` == null).implies(this != null))
```
HT154, line 119: Is this allowed when in a group? If not, then the constraint should express that as well.

packages/com.github.dependabot/v2/Dependabot.pkl
```diff
@@ -242,7 +248,9 @@ class Schedule {
   timezone: Timezone?
 
   /// Specify a valid cron expression for updates
-  cronjob: String?
+  ///
+  /// Required if [interval] is set to "cron".
+  cronjob: String?((interval == "cron").implies(this != null))
```
HT154, line 253: Similarly: is this ever valid when the interval isn't cron?

packages/com.github.dependabot/v2/Dependabot.pkl
bioball: Yeah, I think it is: https://github.com/SchemaStore/schemastore/blob/8012e722834d9fe0b6c99fc2734a7660c29e4046/src/schemas/json/dependabot-2.0.json#L1153-L1163

packages/com.github.dependabot/v2/Dependabot.pkl
bioball: As far as the schema is concerned, there's no disallow condition: https://github.com/SchemaStore/schemastore/blob/8012e722834d9fe0b6c99fc2734a7660c29e4046/src/schemas/json/dependabot-2.0.json#L1177

packages/com.github.dependabot/v2/Dependabot.pkl
bioball: Same; there's no disallow rule for this.

HT154 (approved):

PR #186, opened 2026-05-16
Update AbstractJob.Environment with deployment attribute

joncfoo:
A new attribute named [deployment](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-an-environment-without-creating-a-deployment) can be set to `false` to prevent creating a deployment while still being able to use an environment's secrets and variables.

This change enables that attribute to be set.


HT154 (approved):
Thanks for the contribution!

PR #188, opened 2026-05-27
[com.influxdata.telegraf] Add Docker input plugin

in-the-mood-for-mov:
Hi!

I was trying to generate a configuration file for my Telegraf instance, but there were a few input types not modeled by the library. This PR adds a module for Docker/Podman metrics. Please tell me if this is useful, as I have a few more input types I would like to contribute.

Thank you

### DIFF
--- a/packages/com.influxdata.telegraf/plugins/inputs/DockerInput.pkl
+++ b/packages/com.influxdata.telegraf/plugins/inputs/DockerInput.pkl
@@ -1,0 +1,126 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


packages/com.influxdata.telegraf/plugins/inputs/DockerInput.pkl
```diff
@@ -0,0 +1,126 @@
...
+typealias ContainerState =
+  "created" | "restarting" | "running" | "removing" | "paused" | "exited" | "dead"
+
+typealias StorageObject = "container" | "image" | "volume"
+
+/// Endpoint, either `tcp://<ip>:<port>` for TCP, `unix://<path>` for socket or
+/// `ENV` for using environment variables (i.e., docker-machine)
+///
+/// Default: `unix:///var/run/docker.sock`
+endpoint: String
```
HT154: When we model external config systems in Pkl, fields that are defaulted in-application are encoded as nullable:

```suggestion
endpoint: String?
```

packages/com.influxdata.telegraf/plugins/inputs/DockerInput.pkl
```diff
@@ -0,0 +1,126 @@
...
+
+/// Container labels to exclude as tags, accepting wildcards.
+///
+/// Default: `[]`
+docker_label_exclude: Listing<String>?
+
+/// Which environment variables should we use as a tag
+///
+/// Default: `[]`
+tag_env: Listing<String>
```
HT154:
```suggestion
tag_env: Listing<String>?
```

HT154 (changes requested):
Thanks for the contribution! We'll happily accept this and any other PR adding coverage for other telegraf plugins.

Just a couple small issues to fix before this can merge.

HT154 (approved):
Looks good, thanks!

PR #193, opened 2026-05-29
[com.github.actions] Add new action versions

HT154:
This adds the latest major versions for actions in the `actions` org.

Also: consolidate gradle junit versioning since junit and the junit platform launcher now use the same numbering.


packages/com.github.actions/github/codeql-action/init/v4/CodeqlActionInit.pkl
```diff
@@ -95,15 +88,12 @@ class With {
...
   ///       packages: */*
   ///       token: \$\{{ secrets.GHCR_TOKEN }}
-  /// ```
   ///
   /// The `url` property contains the URL to the container registry you want to connect to.
   ///
-  /// The `packages` property contains a single glob string or a list of glob strings, specifying which packages should
-  /// be retrieved from this particular container registry. Order is important.
-  /// Earlier entries will match before later entries.
+  /// The `packages` property contains a single glob string or a list of glob strings, specifying which packages should be retrieved from this particular container registry. Order is important. Earlier entries will match before later entries.
```
bioball: Revert changes to this file

bioball (approved):
Approving to unblock; one file contains regressions in the doc comments

PR #196, opened 2026-06-03
[com.influxdata.telegraf] Add a few input plugins

in-the-mood-for-mov:
Hi!

This is pull request that adds a few of the simpler input plugins for Telegraf.
* diskio
* kernel
* mem
* processes
* redis
* swap
* system

I'm on the fence about two things :
* Should I split this commit in multiple commits and/or PRs?
* I looked at the existing plugins, some were `open modules` and some were closed. I decided to `open` those that were non-trivial to let the user add configurations that could be out of sync. Maybe I should `open` all of them instead?

Thank you

### DIFF
--- a/packages/com.influxdata.telegraf/plugins/inputs/RedisInput.pkl
+++ b/packages/com.influxdata.telegraf/plugins/inputs/RedisInput.pkl
@@ -1,0 +1,68 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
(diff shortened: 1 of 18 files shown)


packages/com.influxdata.telegraf/plugins/inputs/RedisInput.pkl
```diff
@@ -0,0 +1,68 @@
...
+  /// followed by its arguments.
+  ///
+  /// For example, `new { "get"; "sample-key" }`.
+  command: Listing<String>?
+
+  /// The field to store the result in.
+  field: String?
+
+  /// The type of the result.
+  type: "string" | "integer" | "float" | Null
```
HT154: Reading the plugin source, it appears that all of these properties are actually required:

```suggestion
  /// The command to send to the Redis server, given as the command name
  /// followed by its arguments.
  ///
  /// For example, `new { "get"; "sample-key" }`.
  command: Listing<String>

  /// The field to store the result in.
  field: String

  /// The type of the result.
  type: "string" | "integer" | "float"
```

HT154 (commented):
Thanks! Just one issue in the redis input.

RE: your questions:
* I have no objection to this being one PR; it's of a reasonable size and complexity to review in one go.
* These should be closed modules. I'm not 100% sure why some of the existing ones are `open`, but it's likely historical (and would be a breaking change to fix).

in-the-mood-for-mov:
Thank you for the review, I updated the PR. One question I had that I forgot to ask : for properties such as `RedisInput.Servers`, I considered using a `Listing<Uri>?` (with `URI` from `pkl.experimental.uri`, not the type alias in base), but decided not to because it would add a dependency. What do you think?

bioball:
Yeah, it makes sense to use what's available in the stdlib rather than `pkl.experimental.uri` (that module provides a parser and a structured way to define URIs).

bioball (approved):
LGTM, this is great, thank you!

PR #208, opened 2026-07-01
Bug fix for when a non-nullable property is missing

brimworks:
Without this change, `deepToTyped.attemptApply()` will throw an error if a field is missing that is not marked as nullable when it should be returning a `ConversionFailure` and not throwing.


bioball (approved):
LGTM, thanks for the fix!

PR #224, opened 2026-07-10
Add typealias to represent networks in strict notation

jodersky:
These type aliases require that only the prefix bit of the CIDR notation be set, and all host bits be zero. This is mainly done to enforce style in network configuration but can also help catch errors early.


bioball:
Thanks for the PRs! Can you combine this with #223 and #222? Also, the way this repo works is that every update to a package turns into a new release of that package. So, you'll need to bump the version of this package in this same PR, and also the dependent packages (the build will tell you what dependent versions to bump after you bump this version and then run `./gradlew check`.

jodersky:
Will do! I wasn't sure how you like reviewing changes, and the multi-commit review story on github is suboptimal IMO, hence I split the commits. I'll submit them as one PR with the additional spotless checks and version bump applied.

PR #225, opened 2026-07-10
net: subnet helpers

jodersky:
This pull request combines changes from #222, #223 and #224 (in individual commits, reviewable independently).

The changes proposed are:

- add a `subnet()` method to IPv4Network and IPv6Network, which can compute a subnet directly, without enumerating all
- add `containsNetwork()` and `overlaps()` methods to check subnet relationships between networks
- add a "strict" version of the IPCIDRStrings, which require host bits to be zero
- bump the version from 1.2.1 to 1.3.0, since new features are added

### DIFF
--- a/packages/pkl.experimental.net/net.pkl
+++ b/packages/pkl.experimental.net/net.pkl
@@ -374,7 +374,12 @@ class IPv4Network {
       + ".in-addr.arpa"
 
   /// Calculate all subnets of this network with prefix [target].
+  ///
   /// For example, given IPv4Network("10.53.120.0/21").subdivideTo(24), it outputs 8 /24 networks
+  ///
+  /// Note: this enumerates all subnets eagerly. If the target prefix is much larger than this
+  /// network's prefix, this can take a long time! See [subnet()] for a way to compute a single
+  /// subnet.
   function subdivideTo(target: UInt(isBetween(0, bitWidth) && this >= prefix)): Listing<IPv4Network> =
     if (prefix == target)
       new { self }
@@ -385,6 +390,24 @@ class IPv4Network {
           base = new { repr = self.base.repr + 1.shl(bitWidth - self.prefix - 1) }
           prefix = self.prefix + 1
         }.subdivideTo(target)
+      }
+
+  /// Compute the [n]th subnet of this network with prefix [target].
+  ///
+  /// For example, given `IPv4Network("10.53.120.0/21").subnet(24, 3)`, it outputs
+  /// `IPv4Network("10.53.123.0/24")`.
+  function subnet(
+    target: UInt(isBetween(0, bitWidth) && this >= prefix),
+    n: UInt32
+  ): net.IPv4Network =
+    if (1.shl(target - prefix) <= n)
+      throw(
+        "subnet number \(n) exceeds maximum possible between prefix /\(prefix) and target /\(target)"
+      )
+    else
+      new net.IPv4Network {
+        base = new net.IPv4Address { repr = firstAddress.repr + n.shl(bitWidth - target) }
+        prefix = target
       }
 
   function toString(): String = "\(base.toString())/\(prefix)"
@@ -447,6 +470,10 @@ class IPv6Network {
   /// Calculate all subnets of this network with prefix [target].
   ///
   /// For example, given `IPv6Network("2620:149:a:960::/61").subdivideTo(64)`, it outputs 8 /64 networks
+  ///
+  /// Note: this enumerates all subnets eagerly. If the target prefix is much larger than this
+  /// network's prefix, this can take a long time! See [subnet()] for a way to compute a single
+  /// subnet.
   function subdivideTo(target: UInt(isBetween(0, bitWidth) && this >= prefix)): Listing<IPv6Network> =
     if (prefix == target)
       new { self }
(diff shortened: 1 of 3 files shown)


packages/pkl.experimental.net/net.pkl
```diff
@@ -332,22 +378,62 @@ class IPv4Network {
...
+    new IPv4Address { repr = base.maskHi(prefix) }.toString()
 
   /// Return true if this network contains [ip].
-  function contains(ip: IPv4Address): Boolean = firstAddress.repr <= ip.repr && ip.repr <= lastAddress.repr
+  function contains(ip: IPv4Address): Boolean =
+    firstAddress.repr <= ip.repr && ip.repr <= lastAddress.repr
+
+  /// Return true if [other] is a subnet of this network, i.e. every address in [other] is also in this network.
+  function containsNetwork(other: IPv4Network): Boolean =
+    self.contains(other.firstAddress) && self.contains(other.lastAddress)
```
HT154: Nit: no need to use `self` since implicit `this` gives you what you want. The `self` property is captured above from `this` above so that it can be referenced in scopes where `this` has a different meaning.

```suggestion
  /// Return true if [other] is a subnet of this network, i.e. every address in [other] is also in this network.
  function containsNetwork(other: IPv4Network): Boolean =
    contains(other.firstAddress) && contains(other.lastAddress)
```

packages/pkl.experimental.net/net.pkl
```diff
@@ -332,22 +378,62 @@ class IPv4Network {
...
+
+  /// Return true if this network and [other] share at least one address.
+  function overlaps(other: IPv4Network): Boolean =
+    self.containsNetwork(other) || other.containsNetwork(self)
```
HT154, line 395: This behavior doesn't appear to match the doc comment. It returns:
* `true` if "every ip of `this` IN `other` OR every ip of `other` IN `this`"
* `false` if "any ip of `this` IN `other` OR any ip of `other` in `this`"

I think this is sufficient:
```suggestion
  /// Return true if this network and [other] share at least one address.
  function overlaps(other: IPv4Network): Boolean =
    contains(other.firstAddress) || contains(other.lastAddress) || other.contains(firstAddress) || other.contains(lastAddress)
```

Cases that need test coverage:
* Equality: `this == other`
* Full containment:
  * `this.firstAddress <= other.firstAddress && this.lastAddress >= other.lastAddress`
  * `other.firstAddress <= this.firstAddress && other.lastAddress >= this.lastAddress`
* Partial overlap:
  * `this.firstAddress < other.firstAddress && this.lastAddress >= other.firstAddress && this.lastAddress <= other.lastAddress`
  * `other.firstAddress < this.firstAddress && other.lastAddress >= this.firstAddress && other.lastAddress <= this.lastAddress`

packages/pkl.experimental.net/net.pkl
```diff
@@ -128,26 +135,46 @@ typealias IPv4CIDRString = String(matches(Regex(#"\#(net.ipv4String)/[0-9]{1,2}"
 // language=RegExp
 typealias IPv6CIDRString = String(matches(Regex(#"\#(net.ipv6String)/[0-9]{1,3}"#)))
 
+/// A string that contains a strict IPv4 network, i.e. an IPv4 CIDR block whose address has no bits
+/// set beyond its prefix.
+///
+/// For example, `"10.0.0.0/24"` qualifies, but `"10.0.0.1/24"` does not, since the latter has host bits set.
+typealias IPv4NetworkString =
+  String(this is IPv4CIDRString && let (n = net.IPv4Network(this)) n.base == n.firstAddress)
```
HT154: Use of `&&` in type constraints is usually an anti-pattern. Multiple constrains should instead be separated by `,`, which yields better error messages.


```suggestion
typealias IPv4NetworkString =
  String(this is IPv4CIDRString, let (n = net.IPv4Network(this)) n.base == n.firstAddress)
```

There are a few instances of this pattern in new code, including the `subnet` methods further down.

HT154 (changes requested):
Thanks for the PR!! A couple things need addressing before this can merge.

packages/pkl.experimental.net/net.pkl
jodersky: I agree that the doc comment is misleading. In fact, what I want to say is if either this network is a subnet of the other or vice versa.
If they share at least one address then that is in fact a consequence of them being subnets (i.e. sharing prefixes)

jodersky:
Thanks for the review. I addressed your comments in the latest changes. I think the main noteworthy thing to recheck is the subnet-checking logic. Previously I relied on using first and last addresses, which is technically correct but quite indirect. The new logic uses prefix comparison which I think is the more direct way to check for subnets.

packages/pkl.experimental.net/net.pkl
```diff
@@ -332,22 +378,61 @@ class IPv4Network {
...
+    firstAddress.repr <= ip.repr && ip.repr <= lastAddress.repr
+
+  /// Return true if [other] is a subnet of this network, i.e. other's prefix starts with this
+  /// network's prefix.
+  function containsNetwork(other: IPv4Network): Boolean =
+    prefix <= other.prefix && firstAddress.repr == other.base.repr.and(other.base.maskHi(prefix))
+
+  /// Return true if this network overlaps with the other, i.e. either [other] is a subnet of this
+  /// network or vice versa.
+  function overlaps(other: IPv4Network): Boolean =
```
HT154: I think this doc comment could be clearer and more concise:
```suggestion
  /// Return true [other] is a subnet of this network or vice versa.
  function overlaps(other: IPv4Network): Boolean =
```

The method name also is not super clear without the doc context, which isn't ideal, but I don't have any better suggestions myself.

HT154 (approved):

PR #231, opened 2026-07-28
[inflect] Add package

HT154:
This PR rescues the partial port of [pluralize](https://github.com/plurals/pluralize) from `org.json_schema.contrib`'s internals, implements pluralization (only singularization was ported previously), adds a lot of missing coverage, fixes a few bugs, makes the module's rules extensible, and adds tests (also ported).


packages/inflect/inflect.pkl
```diff
@@ -0,0 +1,310 @@
...
+    .reverse()
+    .toMap((it) -> Regex("(?i)\(it.key)"), (it) -> it.value)
+
+/// Singularize (unpluralize) a word.
+function singularize(word: String): String =
+  inflect(word, irregularSingularizationRules, _singularizationRules)
+
+/// Pluralize a word.
+function pluralize(word: String): String =
+  inflect(word, irregularPluralizationRules, _pluralizationRules)
```
bioball: [nit] since this is the actual API, let's move this to the top

packages/org.json_schema.contrib/internal/ModuleGenerator.pkl
```diff
@@ -28,7 +28,7 @@ import "@syntax/TypeNode.pkl"
 import "@uri/URI.pkl"
 
 import "../ref.pkl"
-import "singularize.pkl"
+import "@inflect/singularize.pkl"
```
bioball: I'd expect this to produce an error in CI; if not, we're missing some test coverage

bioball (approved):
Some comments, but LGTM pending CI

PR #232, opened 2026-07-28
[inflect] Fix double pluralization

HT154:


packages/inflect/inflect.pkl
```diff
@@ -268,7 +268,8 @@ local _singularizationRules: Map<Regex, String> =
 /// Rules for pluralizing a word.
 // language=regexp
 hidden pluralizationRules: Mapping<String(isRegex), String> = new {
-  [#"s?$"#] = "s"
+  [#"$"#] = "s"
+  [#"s$"#] = "s"
```
HT154, line 272: For some reason this wasn't matching greedily, which caused input like `settings` to be double pluralized as `settingss`. Splitting these into two separate rules guarantees that these get evaluated correctly.

bioball (approved):

